PR #32, opened 2016-12-23
Attempt to resolve #30 by allowing AVTs in extension attributes

### LINKED ISSUE #30: Attribute value templates on extension attributes
The specs currently says nothing about AVTs on extension attributes.
Some of these attributes are evaluated while compiling the pipeline (e.g. cx:depends-on), some other may be evaluated while running the pipeline (e.g. cx:message).
AVTs are possible on the latter, but not on the former. To ban uses of AVTs in all cases would be to restrictive.
How do we make the difference?

ndw:


ndw:
Per the 8 Feb 2017, amend to include a note that implementations are free to forbid AVTs in extension attributes if they wish.

ndw:
I believe I've updated this PR appropriately.

PR #91, opened 2017-01-29
Support RelaxNG Compact Syntax media type in <p:validate-with-relax-ng>

### LINKED ISSUE #90: Require processor to support RelaxNG Compact Syntax media type in <p:validate-with-relax-ng>
Currently a processor is only required to use RelaxNG Compact Syntax if the source port document has a `text` media type. The compact syntax has its own `application` media type, `application/relax-ng-compact-syntax`.

dmj:
Fix #90


ndw:
Thank you.

PR #94, opened 2017-03-27
Changed text in 2.7 XPaths in XProc

xml-project:
to XPath 3.1 static and dynamic context items

### DIFF
--- a/langspec/xproc11/xproc.xml
+++ b/langspec/xproc11/xproc.xml
@@ -884,7 +884,7 @@ are visible.</para>
     <section xml:id="xpath-context">
       <title>XPaths in XProc</title>
 
-<para>XProc uses XPath as an expression language. XPath expressions
+<para>XProc uses XPath 3.1 as an expression language. XPath expressions
 are evaluated by the XProc processor in several places: on compound
 steps, to compute the default values of options and the values of
 variables; on atomic steps, to compute the actual values of options.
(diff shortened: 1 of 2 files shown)


ndw:
If you restart the Travis build, does it work?

langspec/xproc11/xproc.xml
```diff
@@ -1027,7 +1030,7 @@ refer to that variable will raise an error.</para>
               </listitem>
             </varlistentry>
             <varlistentry>
-              <term>Base URI</term>
+              <term>Static base URI</term>
               <listitem>
                 <para>The base URI of the element on which the expression occurs.</para>
```
ndw: I don't think this is the static base URI; I think this is the "dynamic" base URI of the actual element in the document.

langspec/xproc11/xproc.xml
```diff
@@ -1196,7 +1258,7 @@ readable port then the context node is undefined.</para>
...
```
ndw: Ditto here, re: base URI.

ndw:
I made a couple of review comments. Aside from those, it looks fine to me.

langspec/xproc11/xproc.xml
xml-project: I used the same term as in XPath 3.1 specs. They say "Static Base URI", but of course we are free to choose an other term.

langspec/xproc11/xproc.xml
xml-project: I used the same term as in XPath 3.1 specs. They say "Static Base URI", but of course we are free to choose an other term.

ndw:
Ok. I have some reservations about that definition, but I agree you've just brought it in line with 3.1.

PR #96, opened 2017-04-05
Fix references to 1.1

ndw:
This PR fixes several references to "1.1" from the prose of the specification.

It also attempts to address editorial credits.


ndw:
Fixed now, I think. Thanks, Gerrit.

Issue #101, opened 2017-04-28
XProc functions only for processor evaluated expressions?


xml-project:
In [2.8 XPath extension functions](http://spec.xproc.org/master/head/xproc30/#xpath-extension-functions) the specs say:
"The XProc processor must support the additional functions described in this section in XPath expressions evaluated by the processor." [Test](http://tests.xproc.org/tests/required/err-d0023-004.xml) raises an error, if 'p:system-property' is used in p:split-sequence, so reads "**only** in XPath expressions evaluated by the processor."

Does it still make sense that only expressions evaluated by the processor (as opposed to expressions evaluated by the steps themselves) have to support these functions? Since we do not support legacy of XPath 1.0 any more, I think we should say, that a processor is required to support these functions for   XPath expressions evaluated by steps as well.

See also #66 for the advantages of having p:document-properties() in p:split-sequence.

ndw:
I favor keeping the restriction for a few reasons:

1. Suppose the implementation of a step isn't an integral part of the XProc pipeline. Suppose I have a web service that can perform XSLT transformations; can I not use that as the implementation of my XSLT step?
2. Most of the extensions are either going to be pointless in the step context, because they return fixed values, or they're going to add a burden to the step implementation to support things that may not be easy.

Consider p:step-available. If that returns "false" irrespective of its argument in a step context (because in what sense is a step "available" in a step context?), what value does it add? If it is supposed to provide, for the sake of argument, "the same answer that the pipeline processor would have given if it was evaluating the function at the point in the pipeline where this step occurs", then the pipeline processor has to provide some mechanism in it's API for step implementors, to find that answer.

The former seems pointless. The latter, burdensome.

xml-project:
I see your point with (1) and I also agree that it may be pointless to use the functions in most cases. 
But:
- @rdeltour in #66 shows very impressive that is **very useful** to be able to use p:document-properties in p:split-sequences. So from my point of view we have to look for another way to solve the problem exposed in #66.
- The specs say that the functions must be visible for expressions evaluated by the context. They do not state, that they **must not** be visible for expressions evaluated by steps. This is only enforced by the test suite.

xml-project:
After our discussion on last weeks telcon and some more thinking about it, I agree that we can not make sure that every step which receives an XPath expression will be able to evaluate XProc's additional functions, e.g. a web service or some third party software encapsulated in an XProc step.

**Proposal**:
[...]

PR #105, opened 2017-05-12
Attempt to fix issue #101

xml-project:


xml-project:
I think the "must not" restriction is what we agreed upon, when discussing this point: See my proposal at the end of the discussion for #101 . I think "must not" is the correct formalisation for "not visible". 

May be we need another loop: Closing pull request and changing label for issue!

(The described behaviour is already enforced by the test suite for XProc 1.0, see test 
http://tests.xproc.org/tests/required/err-d0023-004.xml)

PR #122, opened 2017-08-11
Refactor cleanup

ndw:
For reasons I don't recall, the steps document wasn't actually being validated.

As a consequence, the "x" I left myself at the point where the xi:include was needed to include the serialization options, went undetected and I never fixed that.

And there were some unnecessary namespace declarations.

All trivial fixes, I believe.


ndw:
(Unless we request otherwise in a PR, I think we should take approval by any one of the approvers as sufficient to merge.)

gimsieke:
Yes, we already agreed on that.

ndw:
Right. Just ignore me. My brain is mush these days.

gimsieke:
Then I will disapprove all further pull requests from your part until you present an MD’s attestation that your brain has left the state of mushiness.

ndw:
And in one brief moment, my open source developer career comes to an abrupt end. :-P

PR #126, opened 2017-08-27
Attempt to fix #125

### LINKED ISSUE #125: Error code for defect p:try
We currently have no error code for a p:try without a p:catch (or without a p:group).
Processors behave differently: XMLCalabash XS0027, MorganaXProc XS0044.
Both do not really fit. 

Proposal:
(a) Introduce a new static error XS1012, stating that a p:group and a p:catch is required for a p:try.
(b) Introduce a generic static error message XS1011 stating that it is a static error, if a non-step multi-container step has not the child steps required. This error could be used for p:try and also for an empty p:choose

xml-project:
Adding text for defective p:try and introducing XS1012.

### DIFF
--- a/langspec/xproc30/xproc.xml
+++ b/langspec/xproc30/xproc.xml
@@ -3535,6 +3535,10 @@ the recovery (or “catch”) pipelines are identified with
 <tag>p:catch</tag> elements. The <tag>p:finally</tag> pipeline always
 runs after the <tag>p:try</tag>.</para>
 
+<para><error code="S1012">It is a <glossterm>static error</glossterm> 
+if a <tag>p:try</tag> does not have exactly one <tag>p:group</tag> and 
+at least one <tag>p:catch</tag>.</error></para>
+    
 <para>The <tag>p:try</tag> step evaluates the initial subpipeline and,
 if no errors occur, the outputs of that pipeline are the outputs of
 the <tag>p:try</tag> step. However, if any errors occur, the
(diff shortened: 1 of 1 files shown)


xml-project:
Are you thinking of
<p:try>
   <p:group>... </p:group>
   <p:finally> ... </p:finally>
</p:try>?
To make sense of this construct (apart from commenting out p:catches for debugging purposes) the finally block needs to be executed in any case: when p:group raises an error, finally needs to be executed before the error bubbles up. Is this your point? This would be a useful behaviour even if there are catch elements, because no element may catch the error raised.
Is it this, you have in mind?

gimsieke:
I didn’t have anything in mind with respect to `p:finally`. I just noticed the [editorial note](http://spec.xproc.org/master/head/xproc30/#ednote-finally). I assume Norm put it there. 
Thinking about it now, I can imagine use cases both for `p:finally` in itself and also for a `p:finally` without a preceding `p:catch`. As you say, before the error bubbles up, `p:finally` can clean up temporary files and the like. So maybe we should stipulate that there be at least one `p:catch` or a `p:finally`. It might need further discussion.

xml-project:
Added text to the editor's note on p:finally as requested by @gimsieke 
Merge?

gimsieke:
Yes. I wonder to which (or rather, whose) repo the changes will go if I push the green button “Merge pull request“. If you feel comfortable merging it yourself, please go ahead.

xml-project:
On the top of this page it says: 	xml-project  wants to merge 3 commits into xproc:master
This looks right to me, is it not?

gimsieke:
Yes, it does. Should I push the button? (I’m just a bit cautious because last time around, the changes were merged into Norm’s repo and we never found out why.)

xml-project:
Do it! Last time I merged, it went to the correct repo. So we could test your settings this time.

Issue #129, opened 2017-09-01
What about @p:expand-text


ndw:
Suppose you want to disable expand-text inside an inline? Should we allow @p:expand-text on non-XProc elements? Should it be removed from the generated inline?

xml-project:
I would say: yes and yes!
Having this attribute is not a big burden on the implementation side, but makes node construction a lot easier (even if it is a very special case).
In a typical use case the @p:expand-text is not wanted on the produced node, so it should be removed for the generated node. If we copy it to the inline, a p:delete would be required to remove the attribute. So it is easier for users not to copy the attribute.

Can anyone think of a use case, where you want @p:expand-text to be copied? If there is a use case, obviously the answer to the second question must be "no".

gimsieke:
I see potential issues in two cases: 

1. When someone binds [another namespace URI to the `p` prefix](https://github.com/jpcs/transform.xq/blob/master/lib/patterns.xq#L9) and happens to use an attribute `p:expand-text` in that namespace in their documents. 
1. When someone wants to supply an XProc pipeline, by means of `p:inline`, to, for example, `p:run`, and this pipeline contains another `p:inline` where `@p:expand-text` should be passed verbatim.

The first case can be checked for by looking at the actual namespace URI that is associated with the `p` prefix in the inline document. For text expansion, the processor should only consider (and later remove) `@p:expand-text` if `p` is bound to `http://www.w3.org/ns/xproc`. 

For the (improbable?) second case, we can require the processor to pass `@p:expand-text` verbatim when it has an ancestor `p:inline` (with `p` bound to `http://www.w3.org/ns/xproc`) within the inline document.

ndw:
1. Indeed, the prefix is irrelevant, p:expand-text is just a shorthand for the expand-text attribute in the p: namespace.

2. I suggest we ignore this case in the short term. In the *very* rare case that someone wants to do this, they can always use an add-attribute step and a little complexity to put it back.

PR #131, opened 2017-09-02
Fix 129

ndw:
Here's my attempt to address #129.

### DIFF
--- a/langspec/xproc30/xproc.xml
+++ b/langspec/xproc30/xproc.xml
@@ -4944,12 +4944,11 @@ undefined.</para>
 If it does, the specified value is used as the expand text value
 for the descendants of the element on which it occurs. The
 <tag class="attribute"> will not appear in the inline content.</tag>
-<para>
+</para>
 
-(A <tag class="attribute">p:expand-text</tag> attribute on an element in the
+<para>(A <tag class="attribute">p:expand-text</tag> attribute on an element in the
 XProc namespace is an error. If it occurs on other elements that are not
 descendants of <tag>p:inline</tag>, the attribute has no effect.)</para>
-</para>
 
 </section>
 
--- a/langspec/xproc30/xproc.xml
+++ b/langspec/xproc30/xproc.xml
@@ -4943,12 +4943,23 @@ undefined.</para>
 <emphasis>in the XProc namespace</emphasis>), may appear on an element.
 If it does, the specified value is used as the expand text value
 for the descendants of the element on which it occurs. The
-<tag class="attribute"> will not appear in the inline content.</tag>
+<tag class="attribute">p:expand-text</tag> will not appear in the inline content.
 </para>
 
-<para>(A <tag class="attribute">p:expand-text</tag> attribute on an element in the
-XProc namespace is an error. If it occurs on other elements that are not
-descendants of <tag>p:inline</tag>, the attribute has no effect.)</para>
+<!--feature="p-inline-no-nesting"-->
+<para>If a <tag>p:inline</tag> occurs as
+a descendant of another <tag>p:inline</tag>, any <tag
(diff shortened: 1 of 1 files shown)


langspec/xproc30/xproc.xml
```diff
@@ -4932,10 +4932,24 @@ as defined below, are removed:</para>
...
+<tag>p:inline</tag> is evaluated as a <glossterm>text value
+template</glossterm>. The context node for these expressions is
+undefined.</para>
+
+<para>Within the descendants of <tag>p:inline</tag> the attribute
+<tag class="attribute">p:expand-text</tag> (i.e., the attribute <code>expand-text</code>
+<emphasis>in the XProc namespace</emphasis>), may appear on an element.
+If it does, the specified value is used as the expand text value
+for the descendants of the element on which it occurs. The
+<tag class="attribute"> will not appear in the inline content.</tag>
```
gimsieke: → `<tag class="attribute">p:expand-text</tag>` will not appear in the inline content.

langspec/xproc30/xproc.xml
```diff
@@ -4932,10 +4932,24 @@ as defined below, are removed:</para>
...
+</para>
+
+<para>(A <tag class="attribute">p:expand-text</tag> attribute on an element in the
+XProc namespace is an error. If it occurs on other elements that are not
+descendants of <tag>p:inline</tag>, the attribute has no effect.)</para>
+
```
gimsieke: Suggestion:
If it occurs below another <tag>p:inline</tag> within the topmost <tag>p:inline</tag>, that is, within an inlined pipeline, it will not immediately be used as the expand-text value of the element on which it occurs and it will appear on the inline content.
(It’s better to spell out known fringe cases in adavance. We could alternatively note that people should avoid `<tag class="attribute">p:expand-text</tag>` within inlined pipelines because what happens then is implementation-dependent.)

xml-project: @gimsieke I get (and support) the idea of your suggestion (treating p:inline as a descendant of a p:inline differently), but I can not see how this is reflected by the suggested prose in the specs. Could you please elaborate your idea.

Since the problem with @expand-text we dealing with here, only arises when one wants to construct an XProc pipeline, may be we should mark this and not bother more typical use cases. 
Could we solve this problem in a more transparent way if we abandon the idea of @expand-text having a boolean value? When we say it has three values we could cover the special case with @expand-text="pipeline-construction" (for something like that) to explicitly mark the special case.

gimsieke: I don’t like 'true', 'false' and something else because it looks like a boolean but it isn’t. There’s no precedent to this in XProc AFAIK, therefore it may come as a surprise to authors.
Apart from this, let me try to phrase it differently.
“If `p:inline` contains an XProc pipeline that in turn contains another `p:inline`, any `@p:expand-text` within the inner `p:inline` is treated like an ordinary attribute; that is, neither will its value influence text expansion nor will it be removed from the inlined XProc pipeline.”

xml-project:
Ok, I got it and I think you are right. I think the rephrase suggestions is much clearer than the original prose. 
We just have to say, what it means that a pipeline "contains on XProc pipeline". But I think that can easily be expressed as "any child of an element with name 'declare-step' or 'pipeline' within the XProc namespace". 
If we do it this way, the special treat is invoke by elements name and it's namespace, so we do not need any special value for @expand-text, but can keep a boolean value. 
Good suggestions!

langspec/xproc30/xproc.xml
ndw: Fixed.

langspec/xproc30/xproc.xml
ndw: Fixed.

PR #150, opened 2017-09-26
Fix document-properties

### LINKED ISSUE #149: Fix document-properties
Somewhere along the line we lost document-properties from p:inline and p:document

ndw:
Close #149


ndw:
Asserting this is non-controversial.

Issue #85, opened 2016-12-31
Legacy from 1.0: What document properties are preserved/removed by steps in the pipeline 


xml-project:
# What document properties are preserved/removed by steps in the pipeline 

>This is an aggregation of the discussion on "What document properties are preserved/removed by steps in the pipeline" ([Issue 178 of the xproc/1.0-specification](https://github.com/xproc/1.0-specification/issues/178))

**Opened by: ndw on 2015-06-11, 14:12h**

## ndw said on 2015-06-11, 14:12h:

No description provided. 

	On 2015-06-11, 14:14h: htInEdin was assigned by ndw.

	On 2015-10-07, 14:25h: ndw added this to the XProc 2.0 Next PWD milestone.

	On 2015-10-07, 15:03h: ndw added the awaiting-proposal label.

xml-project:
Also see #88 for the general question, which document properties are present at all.

ndw:
I think this is going to be a hard question, but I also expect it's one that has to be done on a case-by-case basis.

What I want to say is "remove any property that you cannot be certain still applies". But I fear that's going to mean discarding most properties on almost every step. And that greatly reduces the utility of arbitrary properties.

Let's take a few examples. Consider the document "foo.xml" with two properties,`base-uri=/foo.xml` and `user:bar=37`. 

1. p:identity, both properties flow through unchanged
2. p:xslt, generally speaking, the output from XSLT is something entirely different from the input. So we'd expect the output to have a `base-uri` property reflective of the base URI of the produce document. It's hard to imagine that you could confidently assign the `user:bar` property given that you're producing an entirely new document.
3. p:add-attribute, leaves the `base-uri` property unchanged unless the xml:base attribute was changed! What about `user:bar`? I don't know. Suppose 37 is the count of the number of attributes and now we've added one?
4. impl:rasterize-image. *shrug* Who knows?

I fear we're going to have to have some sort of mechanism for pipeline authors to assert which properties should be preserved on an ad hoc basis. Yuck!

ndw:
We own the default namespace and the p: namespace. Everything else should flow through to the largest extent possible.

Issue #154, opened 2017-09-27
Remove proxy documents section


ndw:
Following the decisions we made in Aachen, the proxy document section needs to be updated/remove.

gimsieke:
Do we state instead that XPath expressions see non-XML documents as empty document nodes (with properties)? Previously it was completely implementation-dependent IIRC, which we thought, at the London workshop, to be unwieldy for pipeline authors and bad for interoperability.

ndw:
Done

PR #156, opened 2017-09-27
How do steps effect document properties?

ndw:
This PR attempts to resolve issue #85 

Even if it's imperfect, it's a start.

### DIFF
--- a/langspec/schemas/xproc.rnc
+++ b/langspec/schemas/xproc.rnc
@@ -261,7 +261,6 @@ Inline =
       inline.attributes,
       attribute expand-text { xsd:boolean }?,
       document-properties.attr?,
-      content-type.attr?,
       attribute encoding { text }?,
       use-when.attr?,
       Any
--- a/langspec/schemas/xproc.rng
+++ b/langspec/schemas/xproc.rng
@@ -483,9 +483,6 @@
         <ref name="document-properties.attr"/>
       </optional>
       <optional>
-        <ref name="content-type.attr"/>
-      </optional>
-      <optional>
         <attribute name="encoding"/>
       </optional>
       <optional>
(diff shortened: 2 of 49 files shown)


ndw:
I thought set properties would set the properties to precisely what you provided. So if you don't provide foo, it goes away.

I suppose we could add an "merge" attribute to say that the set properties should be added to the current properties.

ndw:
I added a merge option to p:set-properties and updated the description.

I updated the description of p:split-sequence to preserve the properties.

I don't think, in general, that preserving the properties on XSLT is a good idea. The XSLT step is much too likely, IMHO, to change the document in ways that make them invalid. We'll can have p:with-document (the dual of p:with-properties) for the cases where you know you want to preserve them.

gimsieke:
Do we have another place in the spec where we state to which extent the `base-uri` document property and what is returned by `fn:base-uri(/)` are kept in sync by the processor?
If there is some syncing, then we can add a note that points to this piece of prose, like “The `base-uri` property might be restored on the output because of `p:xslt`’s default behavior and because of the syncing between the `base-uri` property and an XML document’s base URI.”.

ndw:
More good points, thank you! I made another pass through, assuring that we do specify that content-type and base-uri are always set, and consistent. I also took out the proxy documents section (issue #154) because I was in the neighborhood.

ndw:
Foolishly, perhaps, I resolved a merge conflict in the GitHub UI. If it didn't work, things may be unrecoverable. As it is, it's probably beyond my skill to continue making edits to this PR.

gimsieke:
Then perhaps try to merge it right now? Should further changes be necessary, we can file new issues.

ndw:
Ok. I'm going to call that approval and give it a try :-)

PR #159, opened 2017-09-28
First attempt to fix #147

### LINKED ISSUE #147: Serialisation parameters as map?
We have serialisation parameters in p:store, p:http-request etc. 
Would it not be a good idea change the step signature of these steps, so that we have an option "serialisation" which expects a map?
One important reason is have one variable with the map and not having to write them down over and over. The second reason is that we could use fn:json-doc() to load serialisation parameters from a resource.

ndw:
I removed p:serialization, added a serialization attribute and/or option in the appropriate places, and revised the text.


ndw:
I've cited the most recent serialization spec, I think. And I tried to word it so that implementors can support more parameters if they wish. But yes, let's make issues for the details and work through them.

gimsieke:
I’m not an expert with W3C specs, but it seems to me as if https://www.w3.org/TR/xslt-xquery-serialization-31/ is current and supersedes http://www.w3.org/TR/xslt-xquery-serialization-30/ that you cited.

PR #169, opened 2017-10-01
Fix #162 by renaming error codes sequentially

### LINKED ISSUE #162: Rationalize error codes
In order to make editing more practical, we've been inventing error codes haphazardly. I propose to clean that up by renumbering all of the ones currently larger than the largest 1.0 error so that they're sequential again.

I then propose to create a wiki page listing the error codes. Everytime an editor needs to invent a new error code, he or she should add it to the wiki page, thereby getting a sequential list.

ndw:
I'll update https://github.com/xproc/3.0-specification/wiki/Error-Codes in a moment


ndw:
Note that I removed error-list.xml because it wasn't being built correctly anymore.

PR #171, opened 2017-10-02
Attempt to fix #170

### LINKED ISSUE #170: Static checking of serialization parameters
Current version of specs (http://spec.xproc.org/master/head/xproc30/#serialization) says:
"It is a static error (err:XS1201) if the map contains any property name not recognized as a serialization parameter."

I would like to argue to change this to a _dynamic error_, so we could use variables and construct the map at runtime.

xml-project:
Changed static error to dynamic for unrecognized serialization parameters


xml-project:
Conflict between #169  und #170 - my fault. Will do it again!

PR #176, opened 2017-10-02
Attempt to fix issue 153

xml-project:
Introducing a new static error XS0077 for cases, where the value of an attribute of an XProc element does not satisfy the required type, e.g.
&lt;p:input port="prefix:name" /> (NC-Name required) or
&lt;p:inline expand-text='5' /> (boolean required).

We did not have such an error code in XProc 1.0 but I think we should have it in XProc 3.0. May be we could then remove XS0063 (@version must be a decimal).


ndw:
I thought we had a general error for this, but ok.

PR #183, opened 2017-10-04
Add indication of AVTs to the spec

### LINKED ISSUE #181: Clarify AVTs
On steps, option shortcuts can be AVTs. I think we need to clarify what other attributes are, or are not, AVTs.

ndw:
Close #181 

Also: various formatting and CSS improvements. (That gray background in the syntax summaries finally drove me crazy enough to fix it.)


xml-project:
Open questions to discuss:

- serialization parameters should be AVTs. (I think we already agreed on this, when we say we want to use variables for the maps). The values in the map are only used while executing the pipeline, so they can be marked as AVTs safely.
- document-properties should be AVTs to especially in connection with dynamically determined URI (p:document) and dynamically constructed documents (p:inline expand-text="true")
- I am not sure whether {expand-text} does make sense: Either the inline contains TVT statically, then I typically want to expand them always, or the inline does not contain TVTs, then "true/false" will yield the same result.

ndw:
I think that serialization and document-properties are straight up XPath expressions; you can't create a map any other way. We don't need to make them AVTs. In fact, that just makes things worse. Compare:

    serialization="{ map {{ 'method': 'xml' }} }"

with

    serialization="map { 'method': 'xml' }"

We probably need to identify that somehow in the tableaux though.

xml-project:
Just wanted to be sure, I got your point.
Close this?

Issue #182, opened 2017-10-04
Fix content models


ndw:
Fix subpipeline, fix p:try

ndw:
Also with-option needs collection.

PR #184, opened 2017-10-04
Fix content models

ndw:
Fix subpipeline.
Remove variable from p:try (parallel with removing it from p:choose)
Add collection attribute to p:with-option


ndw:
Close #182

ndw:
I think it's ok if there are extra-grammatical constraints. I still think compound steps should have to contain at least one step, even though the content model would allow just variables.

xml-project:
I agree!
Close?

PR #185, opened 2017-10-05
Make expression semantics implementation defined

### LINKED ISSUE #135: User-defined functions per p:variable?
Aren’t function items permitted as variable values? Consider this example:
```xml
  <p:variable name="count" as="xs:integer" select="4"/>
  <p:variable name="foo" 
              select="function($i as xs:integer) as xs:string+ {
                        for $j in (1 to $i) return 'a'
                      }" 
              as="function(xs:integer) as xs:string+"/>
  <p:message message="{$foo($count)}"/><!-- ('a', 'a', 'a', 'a') -->
```
I don’t know how hard this is to implement. In principle, implementations might just translate `$foo($count)` to an XPath 3.1 expression by looking up the declarations, put them after a `let` statement and have the expression evaluated:
```xquery
  let $count := 4,
      $foo := function($i as xs:integer) as xs:string+ {
                for $j in (1 to $i) return 'a'
              } 
  return $foo($count)
```
If they use an XQuery processor to evaluate expressions, they can also add the type constraints:
```xquery
  let $count as xs:integer := 4,
      $foo as function(xs:integer) as xs:string+ := function($i as xs:integer) as xs:string+ {
                for $j in (1 to $i) return 'a'
              } 
  return $foo($count)
```

ndw:
I propose to close #135 with this PR.

### DIFF
--- a/langspec/xproc30/xproc.xml
+++ b/langspec/xproc30/xproc.xml
@@ -4168,10 +4168,14 @@ the value of the variable.</para>
 
 <e:rng-pattern name="VariableSelect"/>
 
-<para>If a <tag class="attribute">select</tag> expression is given, it
+<para>The <tag class="attribute">select</tag> expression
 is evaluated as an XPath expression using the appropriate context as
 described in <xref linkend="xpath-context"/>, for the enclosing
-<glossterm>container</glossterm>.</para>
+<glossterm>container</glossterm>.
+<impl>The precise details about what XPath expressions are allowed
+(for example, can the expression declare a function) is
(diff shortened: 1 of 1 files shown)


xml-project:
@gimsieke this was my idea too: There is a section (https://www.w3.org/TR/xquery-31/#id-conformance)[Conformance] in the XQuery 3.1 specs (I wonder why there is another definition of conformance for XPath 3.1).
Obviously any XProc implementation has to support maps because we heavily rely on them. But to have a list is a good idea.

ndw:
Some sort of conformance language is definitely needed, but I think this fixes the issue in the short term.

PR #187, opened 2017-10-08
Fix #167 again

### LINKED ISSUE #167: Rethink implicit inlines
I think the concept of implicit inline documents is to narrow at the moment. According to our current concept, 
<p:with-input port="port">&lt;element /></p:with-input> 
is allowed, but having two top level elements as children of p:with-input, p:input or p:output is not allowed. 
As we say "An explicit p:inline is required if the author wants to include top level comments, processing instructions, or whitespace, or if the document element is in the XProc namespace." I do not see that this restriction is necessary.
We could imho therefor say in 5.9.3:
"As an authoring convenience, **any element node**, optionally preceded and/or followed by whitespace, in any namespace other than the XProc namespace, occurs where a p:inline is allowed, it is treated as if it was enclosed within a p:inline element (with no attributes)."

ndw:
An input (or with-input) can contain:

  p:empty

or

  any number of p:document or p:inline elements

or

  any number of other elements

In the p:empty/p:document/p:inline case, you can mix in any number of p:documentation and p:pipeinfo elements.

### DIFF
--- a/langspec/schemas/xproc.rnc
+++ b/langspec/schemas/xproc.rnc
@@ -181,7 +181,7 @@ Input =
       common.attributes,
       inline.attributes,
       use-when.attr?,
-      (((\Empty|(Document|Inline)+)? & (Documentation|PipeInfo)*)|AnyElement)
+      ( ( (\Empty | (Document|Inline)*) & (Documentation|PipeInfo)* ) | AnyElement* )
    }
 
 [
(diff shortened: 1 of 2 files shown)


xml-project:
Why not mix other elements, p:document and p:inline?
`<p:with-input port="source">
     <p:document href="something" />
     <doc />
</p:with-input>`
seems perfectly parsable to me.

ndw:
Well. I guess we could. I think it's a bit confusing, but ...

Issue #196, opened 2017-10-12
Allow parameters on p:document / p:load


ndw:
Given that we've extended p:document and p:load to support non-XML data, I think we should allow `parameters` on each. The allowed parameters are implementation-defined and vary according to the content type being loaded. For example, XPath 3.1 defines several parameters to control various aspects of loading JSON content.

If we make this change, I think that 'dtd-validate' should be a boolean property, not an explicit option on p:load.

xml-project:
Yes, we should have a map option on p:document and p:load. For some content-type we should define standard behaviour (e.g. the JSON-load/parse behaviour should be identical the XPath 3.1), for text/* we might come up with some parameters and probably for RDFs.

dtd-validate -> boolean property in a map: Yes!

ndw:
I agree about standardizing parameters for some media types.

PR #199, opened 2017-10-12
Allow parameters option on p:document / p:load

ndw:
This PR allows a parameters option on p:document and harmonizes p:document and p:load.

I took the liberty of removing the attempt to define `p:load` in terms of `p:http-request` since that doesn't seem as relevant anymore. I also removed `override-content-type` from p:document since that's now superseded by the parameters.


ndw:
Close #196

ndw:
Not "override-content-type", simply "content-type" consistent with the other places where we've said that was the semantic of "content-type".

PR #211, opened 2017-10-17
Fix issue 136 by removing p:pipeline from the spec

### LINKED ISSUE #136: Remove p:pipeline (Was: Change implicit primary port behavior for p:pipeline)
Gioele Barabucci suggested that in order to make it easier for newbies who start with `p:pipeline` and then add ports, they sometimes want to rename the primary input or output ports. Instead of forcing them to switch to `p:declare-step`, maybe we can change p:pipeline that you can declare other ports primary (with other names that 'source' and 'result', but maybe even providing a different declaration for a primary input port called 'source', maybe with a default connection). Declaring primary ports explicitly in `p:pipeline` should override the implicit default ports.

ndw:
Several things in here:

1. I updated the schema so there isn't a `p:pipeline` element anymore
2. I fixed the examples, not simply removing `p:pipeline` but also taking care of translating `p:input` into `p:with-input` and other changes. They're schema valid now, but not necessarily correct :-)
3. I updated the prose of the XProc spec. I struggled with what to do with the section that used to be about `p:pipeline`. Replacing it with the section on `p:declare-step` doesn't seem like a perfect solution because readers are going to come to this pipeline section first and there's *a lot* of detail in the `p:declare-step` section. In the end, I just sort of generalized the prose and left it.
4. I fixed some issues in ancillary files related to editing in Emacs. Meh. I should probably take those out of the repo at some point.

At some point, I'm going to print the spec out and do a wholistic once-through with a red pen. Hopefully we can smooth out rough edges like this at that time.

This PR is my attempt to close #136

### DIFF
--- a/src/main/schema/xproc.rnc
+++ b/src/main/schema/xproc.rnc
@@ -117,7 +117,7 @@ decl.attributes = psvi-required.attr?,
 
 # ============================================================
 
-start = Pipeline | Library | DeclareStep
+start = Library | DeclareStep
  | VocabParam | VocabParamSet | VocabResult | VocabHttpRequest
  | VocabHeader | VocabMultipart | VocabBody | VocabHttpResponse | VocabQuery
  | VocabLine | VocabData | VocabDirectory | Errors | StandardStep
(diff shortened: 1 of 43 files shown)


xml-project:
I would approve this, if somebody asks.

Issue #213, opened 2017-10-17
Nodes as a result of a TVT


xml-project:
Unlike in XSLT in XProc 3.0 the evaluation of text value template may produces may, for example if a variable holding nodes is used or if a function is used, which return nodes.

This is fine, but the current version of the specs leaves two cases out:

Case 1: if the node is an attribute, what will happen?
(1) the attribute is added to the "nearest" element node,
(2) the string value of the attribute is added as text node,
(3) a dynamic error is raised.

Case 2: if the node is a document node, what should happen?
(4) All children of the node are inserted,
(5) an error is raised.

Proposal: I would argue to raise an error in both cases, but I am also open for solution (3) and (4).

ndw:
I propose:

Case 1: If the immediately preceding sibling is a start tag, then add the attribute to that element, otherwise error. So `<foo>{$attr}</foo>` is ok. So is `<foo>{$attr1}{$attr2}</foo>` but `<foo>Hello {$attr}</foo>` is an error.

Case 2: Insert all the children.

xml-project:
I agree to "insert all children" in case two.

For case 1: What about <foo>{$att1}{$att2}</foo>?

The first ok (because it immediately follows), but the second not, because it follows {$att1}. If we really want to allow to add calculated attributes, we should take all.

xml-project:
As far as I can see, there is no guarantee that an XPath expression in a TVT evaluates to a single value. It might be a sequence as well, right?
So the prose for attribute case turn out to be complicated, because $attr may contain a mixed sequence of attributes followed by text and in this case
<foo>{$attr}{$attr2}</foo>
should be an error. Agreed?

ndw:
Yes. Ultimately, in tree terms, an attribute can be added if and only if its parent is an element and it has no preceding siblings (or no preceding siblings that are not attribute or namespace nodes, depending on how you do tree construction).

xml-project:
Ok, thats my reading too. 
I will try to make some prose of it (hopefully avoiding the tree terminology).

xml-project:
After thinking about it: Do we really need to say that `<foo>Hello {$attr}</foo>` is an error? To my reading it is perfectly clear what the developer want to do and it is (at least to my reading) in no way ambiguous. 
The only case I would suggest to result in an error, if I try to add an attribute to the element, whose name is equal to the name of an attribute already present.
What did I miss?

ndw:
[...]

PR #222, opened 2017-10-21
Attempt to fix issue 213

xml-project:
Trying to fix issue #213 
I am not sure, whether the formulation for the attribute case is sufficient. Do we need an example case?

### DIFF
--- a/src/main/xml/xproc/xproc.xml
+++ b/src/main/xml/xproc/xproc.xml
@@ -2067,8 +2067,8 @@ element or the attribute has a preceding sibling, that it not an attribute.</err
 </para>
   
 <note>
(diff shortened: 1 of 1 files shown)


ndw:
How can an XPath expression evaluate to something that isn't an atomic value or a node? What are the other possibilities?

By one reading, this explicitly excludes an AVT/TVT that evaluates to a sequence. Is that intentional?

xml-project:
I did not mean to exclude sequences, but to exclude maps, arrays and functions. May be we should say  

> "It is a dynamic error if the XPath expression in an AVT or TVT evaluates to something to other than a sequence containing atomic values or nodes."

But that might be read as: It has to be a sequence and not a single value. (which is of course also a sequence)

PR #223, opened 2017-10-22
Make maps and arrays expressions

### LINKED ISSUE #197: Treat the `parameters` option specially?
I'm on the fence about this one. Consider:

    <p:load parameters="{ map {{ 'liberal': true }} }" href="file.json"/>

That's kind of tedious and `parameters` *has to be* an expression, there's no string value that you can put in there except an expression. So, like `document-properties`, it might be convenient to say that it is always an expression. That would also mean that it could just be an expression on `p:document` (see #196) where it *isn't* an option.

But on steps, *it is* just an "ordinary" option, so treating it specially is an exception that authors will have to remember.

I think, on balance, the exception is going to be less annoying than the AVT. But I'm not sure.

ndw:
This PR attempts to implement @xml-project 's suggestion that we make map and array shortcuts into expressions directly. (As opposed to my question about making the option named `parameters` special.)

I'm still unsure about how users will feel, but I think this implements the change.

Close #197

### DIFF
--- a/src/main/xml/steps/steps.xml
+++ b/src/main/xml/steps/steps.xml
@@ -93,7 +93,7 @@ these steps may be found in
   </listitem>
 </itemizedlist>
 
-<para>Options values are often expressed using <link
+<para>Option values are often expressed using <link
 linkend="option-shortcut">the shortcut syntax</link>. In these cases,
 the option shortcuts are generally treated as <link
 linkend="value-templates">value templates</link>. However, for options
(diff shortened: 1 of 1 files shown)


src/main/xml/steps/steps.xml
```diff
@@ -93,6 +93,18 @@ these steps may be found in
   </listitem>
 </itemizedlist>
 
+<para>Options values are often expressed using <link
```
gimsieke: Shouldn’t it read “Option values“ or “Options’ values“ instead of “Options values“?

ndw:
Ah yes. Ok. I suggest you add maps, arrays, and functions specifically so that other readers won't be as confused as I was. Otherwise, it's fine by me.

PR #227, opened 2017-10-23
Add a global p:timeout option

### LINKED ISSUE #78: Legacy from 1.0: Consider adding @timeout to control potentially long-running steps
# Consider adding @timeout to control potentially long-running steps 

>This is an aggregation of the discussion on "Consider adding @timeout to control potentially long-running steps" ([Issue 162 of the xproc/1.0-specification](https://github.com/xproc/1.0-specification/issues/162))

**Opened by: Conal-Tuohy on 2015-06-04, 03:52h**

## Conal-Tuohy said on 2015-06-04, 03:52h:

Steps like`p:http-request`,`p:xslt`, and`p:xquery`can take a long time to run, and might usefully have an option to put some constraint on that time. e.g. by setting a timeout in milliseconds. 

`<p:http-request timeout="1000"/>` 

If the step failed to terminate in time, it would generate a dynamic error. 

It's common for HTTP Client APIs to offer an explicit timeout, and XProc implementations could easily delegate a timeout to their underlying HTTP API. For instance: [https://hc.apache.org/httpclient-3.x/apidocs/org/apache/commons/httpclient/params/HttpConnectionParams.html#setConnectionTimeout%28int%29](https://hc.apache.org/httpclient-3.x/apidocs/org/apache/commons/httpclient/params/HttpConnectionParams.html#setConnectionTimeout%28int%29) 

The runtime of a`p:xslt`,`p:xquery`, or`p:exec` step is also potentially unbounded. Especially in the framework of a multi-user application it could be useful to be able to kill off a pathological stylesheet or query which has been running too long. 

	On 2015-06-04, 04:07h: Conal-Tuohy referenced this issue.

## ndw said on 2015-06-11, 13:10h:

This seems like an interesting idea. It's not specific to p:xslt, but applies equally to all steps. 

We'd like to get implementor experience about what kinds of security annotations would be most useful/interoperable. 

	On 2015-06-11, 13:10h: ndw added the defer label.

	On 2015-06-11, 13:11h: ndw added this to the XProc 3.0 milestone.

ndw:
I think this is a straight-forward resolution of the issue. I made two conscious decisions not described in the issue:

1. I named the option `p:timeout` because it seems unfair for global options (especially ones likely to be used rarely) to restrict the option names available to step authors.
2. I made it implementation-defined. I'm not confident that every implementor will be able to control the execution time of every step.

Close #78

### DIFF
--- a/src/main/xml/steps/steps.xml
+++ b/src/main/xml/steps/steps.xml
@@ -144,6 +144,46 @@ version or a subsequent version that is backwards compatible with that
 version. At user-option, they may implement other non-backwards
 compatible versions.</para>
 
+<section xml:id="common-options">
+<title>Common options</title>
+
+<para>All XProc steps <rfc2119>must</rfc2119> accept the option
+<option>p:timeout</option>. The value of the
+<option>p:timeout</option> option <rfc2119>must</rfc2119> be a
+non-negative <type>xs:duration</type> or <type>xs:decimal</type>.
+If it is a decimal, it is interpreted as a duration of that many
+seconds. If no timeout is specified, the timeout
+<rfc2119>must</rfc2119> be treated as infinite.</para>
+
+<para><error code="D0053">It is a <glossterm>dynamic error</glossterm> if a step
+runs longer than its timeout value.</error> <impl>It is
+<glossterm>implementation-defined</glossterm> if a processor supports
+timeouts, and if it does, how precisely and precisely how the
+execution time of a step is measured.</impl>
+</para>
+
+<note>
+<title>Note</title>
+<para>Note that the name of this option is <option>p:timeout</option>
+even on steps in the XProc namespace. This limits the extent to which the
+option restricts the space of available option names for steps.
+(If the option name wasn’t in a namespace, then no step could have an option
+named “<literal>timeout</literal>” which seems unreasonable.)
+</para>
+</note>
+</section>
+
+</section>
+
+<section xml:id="steps">
+<title>Steps</title>
+
+<para>This specification describes both required and optional steps.
+All implementations <rfc2119>must</rfc2119> implement the required steps.
+If an implementation chooses to implement an optional step, it
+<rfc2119>must</rfc2119> be implemented with the semantics described in
+this specification.</para>
+
 <section xml:id="std-required">
 <title>Required Steps</title>
 


xml-project:
I know I am not ask for a review on this, but I like to express my opinion anyway:
(1) The initial request was a timeout for p:http-request, more precisely the time a step would wait for the resource to be delivered. I take "waiting for a resource" to be a much more precise term (at least in some intuitive sense) than "time a step runs".
In the specs there is no concept of "the time a step runs" and I think it should not be in there. The specs should say what is supposed to happen and not limit implementers, how and when its is going to happen.
And: What is counted in here? The time it takes to evaluate the variables used in the option expressions? Not the variables evaluation time, but the option evaluation time? The evaluation of the step binding? And suppose the processor stops the execution and has to do some cleanup: Has this time to be counted in in advance to conform to this feature?
(2) I agree with you that it has to implementation-defined because I do not think that this feature should limit implementers in the architectural choices.
(3) Not only the question whether it implemented or not it implementation-defined, but also [see (1)] what will count as "time the step is running" and how precise this time is taken.
(4) Could there be a test case in the test-suite checking whether a processor conforms to this feature by running a step no longer than x seconds?

Conclusion: I would like to see a time-out feature on p:http-request because it seems useful in interactive XProc pipelines, but I doubt that the generalization to p:timeout really makes sense. It is to vague from my point of view so I would suggest not to have it in the specs. Of course every implementer is free to implement it with an attribute in her own namespace. And I would surely prefer this because it gives a clear hint to developers they using a non interoperable feature.

gimsieke:
As a user, I’d like to be able to specify a timeout also for other steps, for example Schematron validation.
Case in point: Very recently, an inefficient regex made a docx→TEI→EPUB conversion pipeline that usually runs less than 3 minutes run for more than half an hour (see the [Saxon issue](https://saxonica.plan.io/issues/3494?pn=1#change-10372) that I posted last night). This didn’t happen before because the regex was written while we were using Saxon 9.5 (where the regex ran just fine) and because the reported phenomenon didn’t happen 312 times in a book before. So we are definitely interested in sticking a timeout to Schematron validation, the value of the timeout probably also being dependent on the input length, lets say `string-length(/) * 1E-4` seconds, that is, less than 2 minutes for 1 MB of text. In case of a timeout, we’d catch this error, and ideally receive the SVRL that was produced so far. I admit that finishing each step nicely when the timeout occurs will be harder to implement. If we don’t get any output, we can still report this as a timeout error and instruct the end users to ask our staff to look into the issue.
So I think Norm’s suggestions are just fine, and `p:timeout` support for any step will definitely be a selection criterion for transpect’s future XProc processor. (transpect 3.0 will run on both Morgana and Calabash, but we’ll set up most projects with just one processor distro attached as a submodule or external.)
Each processor’s concept of what counts as running time may vary, but if Morgana thinks a step ran for 30 seconds while Calabash’s calculations amount to 32 seconds, this is absolutely fine. We can add a note that running time should be measured as the real time after entering the step, although details are implementation-dependent and measurements can differ, not least because of the overall system load, different concepts of when a step begins to run, etc.

xml-project:
Gerrit, I am not arguing against the idea of having a time-out, but against 
(a) the idea, that it is the same concept as "wait for a web server to respond" and 
(b) that it is something which is specific to the language XProc as opposed to being specific to an implementation of XProc.
The case you made that XMLCalabash might to 30 seconds while MorganaXProc might take 32 seconds is imho an argument for (b), because if someone  uses p:timeout="30" according to Norm's suggestion one processor will behave in accordance with the specs while the other will fail. If we make it a processor specific feature, the implementor can tell which decisions she made to count the actual amount of time. So running 30 seconds on processor A could means something else than running 30 seconds on processor B. This gives cutting-edge XProc developers the opportunity to use "cx:timeout='30' mox:timeout='34'" because they know that MorganaXProc is slower here.
 
Having it in the specs means that '30 seconds' is a standard without any exception, so it will dominate the design of the implementation to be able to match this limit. And I do not see, that this point has the importance to dominate over other design criteria as for example low memory consumption and effective use of machine resources. 

Let me go into some detail to make this point: My current implementation of MorganaXProc 2 is a multi threaded (or parallel) implementation, which makes use of the multiple cores in modern machines. This gives an overall performance improvement but makes it harder to say, how long a step has already run. It might be possible to have some exact timing here, but that would certainly result in some performance reducing overhead. And if there is no way to precisely control the time, a step has already run, I am facing a dilemma: Either return to a monolithic system which is less performant but allows me to control the time already elapsed or having a fast processor which does not conform to the specifications because it does not throw the exception expected after 30 seconds but after 31 seconds or 40 seconds.

And this dilemma makes me asking whether this feature is really so important that we need it in the specs. And, to repeat it: It is something other than "wait for a web server" and I would like to see this feature.

gimsieke:
It might even happen that Calabash takes 27 seconds for one invocation and 32 seconds for the other, depending mostly on system load. The point is that we want to make sure that a step runs no longer than a minute instead of 2 hours when a regex backtracking fails. The actual time depends more on factors outside of the processor’s control than on different implementations / measurement methods. 
We should add a note that this is intended for steps that run longer than approx 15 seconds and that its main purpose is to avert catastrophic failures. Pipeline authors should be aware that it may perfectly happen that the timeout strikes with one invocation and doesn’t strike with the next, or that different implementations measure things differently.
One example for different measurements may be dynamic evaluation. For complex dependencies, a processor might need several seconds to compute the dependency graph. Even if both processors need several seconds, one may choose to include graph calculation in the running time while the other does not. Still, there is absolutely no need to specify processor-dependent timeouts. If you know that a given processor typically reports 50 seconds while the other reports just 30, although they take the same overall time, and you can tolerate both times, you just specify a timeout of, say, 300 seconds, because you only want to protect yourself from accidental really catastrophic running times.
If you want to prepare a case for a test suite, you will probably create a loop that will take minutes on current hardware and add a `p:timeout="1"` (if we implicitly interpret the value as seconds – or do we want xs:dayTimeDuration?), see whether the processor terminates after a few seconds, and take this as an indication that the processor supports `p:timeout`.
Even in a parallelized scenario, it should be possible to look at the real-time clock when a step starts and when it ends, and to subtract both values.
Finally, if the support of `p:timeout` brings with it a general performance penalty, then you can add a flag to disable (or to enable) timeout checks.

xml-project:
> Even in a parallelized scenario, it should be possible to look at the real-time clock when a step starts and when it ends, and to subtract both values.

Of course this is possible, no question. But since the process handling the step is mostly asleep while waiting for the options and bindings to be delivered, one has to wake him up periodically to look on the clock.

> Finally, if the support of p:timeout brings with it a general performance penalty, then you can add a flag to disable (or to enable) timeout checks.

This is not an option imho because whether an implementation is doing precise time measurement and doing no (precise) measurement is an overall design decision. So having such a switch means having two different XProc implementations as far as I can see.

Finally: I we share your relaxed interpretation of how long 30 seconds are (I am not sure we all do), then we can not have an error message saying:

> It is a <glossterm>dynamic error</glossterm> if a step runs longer than its timeout value.

and then add that it implementation defined how good the time measurement of the machine is.

@gimsieke I really see your point but I still not convinced that this is the best way to do it.

ndw:
First, my failure to include @xml-project in the review requests was purely that: a failure. I swear I checked both names. Oh, well. Apologies, Achim.

As to the proposal itself, I don't feel strongly about it. If you look at #78 I think you'll see that what we recorded as the decision at the workshop in Aachen was that all steps would have a timeout attribute. I was just trying to implement that. It looked like low hanging fruit after dinner and a couple of glasses of wine :-)

I assumed that the intent was a step would be terminated after its timeout passed. Looking at at least one of @gimsieke 's responses suggests that perhaps that's not what he has in mind.

My implementation plan is, if a step has a timeout on it, start a thread with a sleep of that length. If the thread wakes up again before the step finishes, throw the exception. That's going to suffer from the vagaries of how precisely Thread.sleep is, but I'm not going to sweat it. I think we're looking for "did this step run more than two minutes" and not "did this step run for more then 31.4159s"

I'd be happy to move this into the "injection" spec as well, FWIW.

ndw:
Agreed, @gimsieke

xml-project:
@ndw You do not have to apologise. I thought it was slip. May be I should have marked my introduction ask a joke.

I am still not convinced by the solution mixing two very different cases. I do not exactly recall our discussion in Aachen but there surely was a consensus to have the opportunity to stop the running of a step if it takes to long.

Will you two please give me the change to develop a proposal which matched Conal's and my interest in controlling the user experience of web applications talking to servers via p:http-request and Gerrit's interest in setting some time limit for the execution of a step because he knows, that something went wrong.

gimsieke:
The only extra that we wished for in an ideal world is a signal to the step: Finish what you are doing, close the result tree and return it. But this is not realistic. Therefore, we are happy with the current proposal, provided that `p:timeout` can be calculated and evaluated at run time. 
This feature is important to us because we’re seeing excessively long-running processes from time to time and most of the time users complain about it. (Sometimes it happens after Saxon updates, sometimes it’s due to unexpected inputs.) We’d prefer the pipeline rather than the users reporting these long-running steps.
As I said, we don’t care about vagueness in the order of several seconds, or even minutes for steps that typically run for minutes. We cannot have perfect reproducibility here anyway. If a steps is terminated because it runs slow due to high system load, all the better. This is also a thing that we want to be notified about.

ndw:
Replaced by #243

PR #233, opened 2017-11-01
Attempt to address selection and content types

### LINKED ISSUE #193: @select and document-properties
The example @ndw provided in his [mail](http://markmail.org/search/?q=list%3Axproc-dev#query:list%3Axproc-dev+page:2+mid:exbrvizz3bbplrjb+state:results) to me raises another point with @select on p:with-input etc.:
Suppose I have a step which accepts say "text/plain" and I have an XML document containing a text node as child of a "section" element. Can I use
`<p:with-input select="//section/text">`
with the XML document as default readable port or context node (if you prefer)?

The problem is at which point we will say, that the document to arrive on the text step does not match the expected content type "text/plain". If we do it before the evaluation @select an error has to be raised because there its "application/xml". If we consider the result of @select (as we do to assess whether its a sequence or not), no error should be raised.

I think you can make up the same kind of example for any XDM defined type as value of @select.

I hope, I made my point clear. If not, please tell me so and I try to provide an elaborated example.

ndw:
This PR attempts to fix #193 

The significant changes are:

* Make the content type of documents created by `select` explicit. If a document is entirely text nodes, it's `text/plain` otherwise it's `application/xml`.
* Remove `err:XD0016`
* I moved the description of `select` before `content-types` in `p:input`

I made a few other editorial fixes.


ndw:
That's an error. Document nodes can't have attribute children.

xml-project:
`That's an error. Document nodes can't have attribute children.`

a) agreed,
b) did not get your point because I was not talking about document nodes at all but about  producing an xs:integer-Sequence with a select.

ndw:
Sorry, I missed the xs:integer() part. I don't think that's an error. I think it's a document node that contains a single atomic value. My proposal says it has a media type of application/xml because it's not a text node. It's not well formed XML, of course, but it's an XDM fragment that you can do XPath expressions over.

The special case for text is just to allow XML documents to be fed into steps that expect text without requiring some kind of intermediate transformation step. If you have a "text:head" step that returns the first 10 lines of a text file, I kind of want this to work:

```
<text:head>
  <p:input port="source" select="string(/)" href="data.xml"/>
</text:head>
```

xml-project:
I like the idea, but I am not sure why we say its application/xml because it is not. I would tend to think, that we should have "application/xml" only for proper XML. Since it is not text, it may be application/octet-stream.
Suppose I have a step bytes:count accepting "application/octet-stream" I could do

` <bytes:count><p:with-input port="source" select="xs:hex-binary(string(/))" href="data.xml" />`

ndw:
I'd be reluctant to say that application/xml (between steps; this isn't about serialization) requires well-formed XML. I'd like to say that, actually, but the reality is that lots of processes can produce XDM instances that aren't well-formed but most XML processes that consume XDMs (xslt, xquery, our own steps, etc.) don't actually care.

There's a slippery slope here: xdm, text, single atomic values, hexbinary, lists of atomic values of different types, hexbinary and base64 encoded etc etc etc

But if you want to draw the line somewhere else, please make a proposal.

xml-project:
I am not sure we have the same understanding of XDM (and of course you are the authority here):
If I do "document{xs:integer(4)}" in XQuery I do not get an a document node containing an atomic XDM-value, but (to my understanding), I get a document node containing a single TEXT NODE with value '4', so the type information is lost.
The same holds if p:xslt or p:xquery return an atomic XDM value and we wrap a document-node around, right?
In one of our discussions in [Prague](https://github.com/xproc/Workshop-2017-02/wiki/Minutes-for-day-one:-7-February-2017) you proposed:
“(1,2,3”) content-type=“x.vnd/xdm-item”
May be this could be a way to do it: We do not wrap a document-node around it, but say its representation is implementation-defined, but it's content-type is "x.vnd/xdm-integer" or (in other cases) "x.vnd/xdm-hexBinary". 
Do not know, whether is it really worth doing it. May be we come back to it, when talking about the p:xquery etc. return XDM values.

PR #243, opened 2017-12-11
Proposal for p:terminate-after (Issue #78)

xml-project:
Counter proposal to resolve issue #78. For semantic reasons I changed "p:timeout" to "p:terminate-after". I will propose to add an option named "timeout" on p:http-request. Having "p:timeout" (a common option) and "timeout" seems to be to error-prone.

@gimsieke @ndw
Due to the previous discussion I propose that we should have a consensus on this before merging.

Proposal is now complete. Forgot to add a note to the previous pull request

### DIFF
--- a/src/main/xml/steps/steps.xml
+++ b/src/main/xml/steps/steps.xml
@@ -171,7 +171,7 @@ the XProc processor's execution strategy, the system load etc., this feature can
 be used as an exact timing tool in XProc. Execution times may vary on different executions 
 of the same pipeline on the same computer and the same XProc processor due to  
 execution contingencies. Developers are advised to calculate the value for p:terminate-after 
-generously, so the dynamic error is raised only in extreme cases.</note>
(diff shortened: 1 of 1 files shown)


ndw:
I'll defer to @gimsieke on the name. I have a preference for @`p:timeout` because that's what this thing is called in other contexts. I don't think it conflicts with `p:http-request` because I think the request timeout belongs in the request, not as a property on the step.

But I won't fuss over @`terminate-after` if that's what y'all want.

gimsieke:
I’d prefer 'timeout', too.
Regarding `p:http-request`, I thought we can do without allowing a timeout on `c:request` but instead use the generic timeout mechanism on the `p:http-request` step.  
I think this can best be discussed in a brief call this or next week if you can spare 20 minutes.

xml-project:
Changed the option's name to p:timeout.

xml-project:
I have added the note on p:http-request as requested by @gimsieke 
Can I merge this now?

PR #244, opened 2017-12-11
Timeout on http request issue #78

xml-project:
Proposal for attributes timeout and fail-on-timeout in c:request of p:http-request

### DIFF
--- a/src/main/schema/xproc.rnc
+++ b/src/main/schema/xproc.rnc
@@ -700,6 +700,8 @@ VocabHttpRequest =
       attribute auth-method { text }?,
       attribute send-authorization { xsd:boolean }?,
       attribute override-content-type { ContentType }?,
+      attribute timeout {xsd:positiveInteger}?,
+      attribute fail-on-timeout {xsd:boolean}?, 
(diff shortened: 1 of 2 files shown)


ndw:
I hate to be argumentative, but why is this on the `p:http-request` step and not part of the `c:request`? It feels to me like it should be part of the request.

gimsieke:
But the timeout, as Achim proposed it, is on `c:request`, not on `p:http-request`. 
I was thinking that we can do without a timeout on the request but instead use the generic timeout mechanism. Processors should be allowed to selectively support the generic timeout only on specific, implementation-defined, steps. This would allow them to support a timeout only on `p:http-request`, for example.
As I suggested in the other comment, maybe we should discuss the whole timeout issue (HTTP request and otherwise) in a hangout.

ndw:
A request timeout is a pretty common feature of http libraries. There are good reasons for sometimes wanting to change it (for example, making it very short if you're just pinging to see if a service is available). Sorry I misunderstood where this was being proposed, Achim.

xml-project:
I have added the note on p:http-request as requested by @gimsieke 
Can I merge this now?

Issue #239, opened 2017-12-02
Introduce attribute to control visibility of steps for an imported pipeline


xml-project:
Recently I had a p:library with four step declarations, where one is called by the other three, but should not be called directly from an importing step.
To give pipeline author more control I propose an optional additional attribute on p:declare-step which is only relevant when the p:declare-step appears inside a p:library. 
The attribute might be called "visibility" (like in XSLT 3.0) and could take the values "public" or "private". If there is no attribute "visibility" on a p:declare-step the step is visible when the p:library is imported by another step. But if "visibility" is "private" the step is only visible inside the p:library it is declared on.
To align the approach with the same attribute name and values as in XSLT seem reasonable to make the use easier for pipeline authors familiar with XSLT. But we might also think about an attribute named "export" with takes a boolean value.

gimsieke:
I think this is a useful proposal. I cannot judge how easy it will be to implement though. My feeling is that it shouldn’t be that hard for any implementation. Some modifications to the part in the spec that is about [visibility](http://spec.xproc.org/master/head/xproc30/#dt-visible) would be due.
I was briefly considering a different approach: Keep everything in libraries visible, but let the pipeline author (i.e., the user of the library) decide which steps they need, like in Perl’s `use` instruction (for example, `use sigtrap qw(SEGV BUS);`). So you could add an attribute called `steps` to `p:import` that accepts a list of QNames for the steps to be imported (import all steps if the attribute is omitted). 
But this approach solves a different (non-) issue. I think you approach is better because library authors know best which implementation details of their library they want to hide.

xml-project:
@gimsieke Which notation would you prefer? 

- visibility="public | private" or 
- export (xs:boolean)

gimsieke:
visibility

ndw:
I vote for visibility as well.

xml-project:
Thanks. I will prepare some prose to cover this case.

xml-project:
Solved with rp #245

PR #245, opened 2017-12-11
Attempt to fix Issue #239

xml-project:
Let's see if I can do something not controversial today:
I introduced attribute "visibility" on p:declare-step and add the prose to explain its use.

Deviating from my original proposal I made it an xs:boolean for two reasons:

1. Doing so, we do not need to have another error which is raised when a wrong value is used for the attribute.

1. I could not figure out how to do this in RNC.

If you like the original proposal better, I will be happy to change the prose if someone changes the rnc (or tells me how to do ("public" | "private").

### DIFF
--- a/src/main/schema/xproc.rnc
+++ b/src/main/schema/xproc.rnc
@@ -392,6 +392,7 @@ DeclareStep =
       name.ncname.attr?,
       attribute type { xsd:QName }?,
       decl.attributes,
+      attribute visibility { xsd:boolean }?,
       common.attributes,
       use-when.attr?,
       (Input|Output|Option|Documentation|PipeInfo)*,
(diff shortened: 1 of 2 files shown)


gimsieke:
If the value is boolean, I’d call the attribute 'public'.

ndw:
We know a lot of XProc users also use XSLT, so I think I'd prefer `visibility` as you earlier proposed.

In the RNC file, something like this should work

    attribute visiblity { "private" | "public" }?

xml-project:
> Am 12.12.2017 um 02:20 schrieb Norman Walsh <notifications@github.com>:
> 
> We know a lot of XProc users also use XSLT, so I think I'd prefer visibility as you earlier proposed.
> 
> In the RNC file, something like this should work
> 
> attribute visiblity { "private" | "public" }?
> 
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub, or mute the thread.
> 

Aligning with XSLT was my first thought too. Will change my PR today.
Thanks for your help.

xml-project:
I have change syntax and prose to "private | public". The RNC is ok (according to Oxygen) but there seems to be a problem with rendering because the quotes around string constants are missing.

PR #264, opened 2018-02-02
Attempt to fix #263

### LINKED ISSUE #263: Error for syntactically invalid sentence types
According to the error wiki XD0049 is raised for syntactically invalid sequence types, but this error is not mentioned in the specs.

xml-project:


ndw:
Note to self: it has to be a dynamic error (not a static error) because it might rely on types that aren't known until runtime. (Assuming, for example, that the compiler doesn't load schemas or that the schemas could change between compilation and runtime.)

This is a good candidate for a dynamic error caught statically, however.

PR #307, opened 2018-02-26
Spec refactor

ndw:
This is a big slug of changes. Essentially it accomplishes these goals:

* It removes random dependencies from the build (`perl`, `make`, `cp`).
* It rationalizes the build system, improves how various artifacts like schemas are constructed.
* Per decisions on Prague, it begins the process of breaking the optional steps into separate specifications.
* It restructures the publication so that there's an Overview spec that points to all the others.
* I created a new "step introduction" spec that describes common features of steps so that they don't have to be repeated in each step spec

It's not supposed to have introduced any substantive changes to the spec prose.


ndw:
All the editors agreed and no one on the list complained, so ...

PR #308, opened 2018-02-28
Make signature of p:store more explicit

xml-project:
After some question already raised I think its better to flag output port "result" explicit as primary="true" to mark the difference to XProc 1.0
And, to be honest, is also a test the new system ;-))

### DIFF
--- a/steps/src/main/xml/steps/store.xml
+++ b/steps/src/main/xml/steps/store.xml
@@ -7,7 +7,7 @@ location of the stored document.</para>
 
 <p:declare-step type="p:store">
   <p:input port="source" content-types="*/*"/>
-  <p:output port="result" content-types="application/xml"/>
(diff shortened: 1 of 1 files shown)


ndw:
But it only has one output port so it *will* be primary by default.
Are you proposing that we make all of the output ports explicit, or just this one?

xml-project:
Yes, I know this rule. But as we had a question about the _change_ on p:store (was primary='false' in XProc 1.0) at Prague and also on the mailing list, I thought it might be a good idea to flag this change explicitly (at least for a while). I do not think it is necessary in the other places where nothing has changed.

Second I wanted to try out the new deployment system you set up.

If you do not like it, feel free to say "no" to the pull-request. I will not insist on this point.

ndw:
No, that's fine. I just wanted to make sure I understood.

PR #314, opened 2018-03-02
Attempt to fix issue 297

xml-project:
Removing reference to p:template from step master
Delete file containing steps description


ndw:
Ok. I had second thoughts about pulling it out. I wanted to check that a p:identity with a p:inline inside it would really be equivalent. But I guess we can put it back if we find we need it.

PR #329, opened 2018-03-17
Clarify the semantics of p:finally, relax restrictions on p:catch

### LINKED ISSUE #306: Clarify semantics of p:try (without p:catch)
After reading the specs on p:try again I think we need to clarify the semantics of a p:try which has no p:catch but a p:finally. 
Two interpretations of behaviour in an error case:
(1) Since no p:catch is specified here but there is a p:try, the error is discarded and p:finally is executed. The output port of p:try are the output ports of the p:group (or of the subpipeline since p:group is not required anymore), but all output ports are empty.
(2) Since no p:catch is stated here, the error has to bubble up to an outer p:catch or the processor. But since there is a p:finally, this subpipeline is executed before the error bubbles up.

I am in favour of interpretation (2) which would be _my_ natural reading of the pipeline, but I am certainly open to arguments.

However we decide: Since we made p:catch optional we need to explain, what p:try-p:finally means.
Any thoughts?

ndw:
This PR
1. Simplifies the content model of p:try.
2. Clarifies that p:finally always runs.
3. Relaxes the restrictions on p:catch a bit. Given p:finally and the possibility of no catches matching, I think I've simplified and improved things.

Close #306

### DIFF
--- a/src/main/schema/xproc.rnc
+++ b/src/main/schema/xproc.rnc
@@ -495,7 +495,7 @@ Try =
       name.ncname.attr?,
       common.attributes,
       use-when.attr?,
-      ((TryGroup, Catch*, Finally?) & (Documentation|PipeInfo)*)
+      ((TryGroup, ((Catch+, Finally?)|(Catch*, Finally))) & (Documentation|PipeInfo)*)
    }
 
 [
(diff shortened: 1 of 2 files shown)


ndw:
Bah. I oversimplified.

ndw:
I'm assuming we're all happy now that I fixed the content model.

PR #331, opened 2018-03-17
Make a single validation step spec

ndw:
Combined the validation steps into a single spec per workshop directions.
Cleaned up a bit of formatting.
Removed some bogus references.


xml-project:
Something seen to be wrong, because Travis says "All checks have failed"?

ndw:
Yes. It builds locally but not on Travis. I'm poking it with a sharp stick now.

PR #360, opened 2018-04-09
Added @tunnel to p:option (description only)

eriksiegel:
Hi. I added some text about the tunneling option to the specification. I don't feel enough at home to make the (probably simple) change to the schema also, so that remains to be done. 
It might also be that my writings are not exact enough for the implementors but it can serve as a starting point?

### DIFF
--- a/xproc/src/main/xml/specification.xml
+++ b/xproc/src/main/xml/specification.xml
@@ -4642,6 +4642,21 @@ default value.</error></para>
 the <tag class="attribute">select</tag> expression makes reference to
 the context node, size, or position.</error></para>
 
+  <para>An option may declare itself tunneling by specifying the value
+    <literal>true</literal> for the <tag class="attribute">tunnel</tag>
+    attribute. If a called step has declared an option as tunneling and
+    the calling step does not specify a value for that option using
+    <tag>p:with-option</tag>, the in-scope value from the calling step
+    is inherited. This allows passing often used options
+    implicitly.</para>
+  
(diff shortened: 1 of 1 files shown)


eriksiegel:
We could make it easy on ourselves (like Tom Jones sings, I can hear him in my head while I write this) and make the rules as strict as we like.

Proposal:
- Tunneling implies required="false". Setting required="true" and tunnel="true"is an error
- You can set a tunnel option by either using an explicit with-option (this takes precedence) *or*, when there is no with-option for this option, having an option  in the calling stylesheet with the same name and tunnel="true"
- If you add a select attribute on option, this simply takes over when the option is not set, like always.

xml-project:
I think this is a good approach, I agree with all three proposals.
Speaking of "easy solutions" (btw. I do not know the song): I doubt whether we need XS0088. If we leave it open, the appearing value, where ever it come from (explicit, tunnelling, default) has to match the type for the option in question, otherwise it's XD0036.
Doing it this way would also allow to cast types which would be prohibited by comparing string because an xs:integer value is castable to xs:decimal, but "xs:integer" is not equal to "xs:decimal".

eriksiegel:
All right, I think the PR was a bit too premature. I reject it. The text needs to be changed anyway *after* we reach full consensus.

xml-project:
@eriksiegel Do you think it would be useful to copy your proposals to the related issue?

eriksiegel:
Yes, I will do this today.

Issue #380, opened 2018-04-25
select on p:with-input / p:input


xml-project:
We have to specify what happens, when `@select` on p:input/p:with-input selects a node:
(1) text document with the text value is created
(2) error

I would tend to (2).

xml-project:
Fixed with pr #381

PR #381, opened 2018-04-25
Attempt to fix 380

xml-project:
Fixing issue #380 by explicitly saying, which results of `@select` are allowed.


xml-project:
@ndw @gimsieke @eriksiegel 
Is there a problem with this PR? I thought it would be pretty uncontroversal.

gimsieke:
The purpose is to exclude attribute and namespace nodes?

xml-project:
Yes, because the specs say that a new document is created by wrapping a document-node around the select result if it is not a document-node.
Then it must be an error if "@select" returns atomic values, maps, arrays, function items, attributes or namespace nodes, right?

gimsieke:
Atomic values, maps, arrays, and function items have already been excluded by reference to “nodes”. But yes, let’s go with explicitly listing the permitted types.
I originally thought we could refer to the selectable nodes using [`node()`](https://www.w3.org/TR/xslt-30/#d7e19631) in the XSLT pattern sense. But this includes only text, element, PI, and comment, not document nodes.

Issue #382, opened 2018-04-29
Error XC0023 and XC0025 at p:insert


xml-project:
I think we should have a look at the prose and errors for p:insert. Currently it is forbidden to match the document node (XC0023), but I can not see why. To me it seems perfectly natural to match the document node and then to insert e.g. a processing node as first child.

Also XC0025 seems strange because a document node has a first and a last child. I think the error must be raised when the document node is matched and "@position" is "before" or "after".

Where is my mistake?

xml-project:
Fixed with pr #383

PR #383, opened 2018-04-30
Attempt to fix 382

xml-project:
Fixing issue #382 by 
(1) allowing document nodes to be matched in general (so now only attribute node and namespace nodes are forbidden), 
(2) allowing document nodes to be matched, when position is first-/last-child.
(3) introducing new error (C0024) which is raised, when document node is matched and position is "before" or "after".


xml-project:
Thanks for your comment @gimsieke. I would say that we should make the clarification you asked for in the general part and not for every step that has a match attribute again. If we do it this way, how are the changes I proposed affected?

xml-project:
@gimsieke created issue #384 for your comment.

PR #396, opened 2018-05-18
upgrade to  calabash 1.1.20-98

innovimax:


ndw:
Seems reasonable to me. Thanks!

PR #412, opened 2018-06-21
Build changes for portability

ndw:
I believe these changes fix the build problems on Windows.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,5 +1,6 @@
 buildscript {
   repositories {
+    maven { url uri('file:/F:/.m2/repository') }
     mavenLocal()
     mavenCentral()
     maven { url "http://maven.restlet.org" }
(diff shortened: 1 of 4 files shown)


eriksiegel:
Thanks! Very busy now, will try to look into this tomorrow.

xml-project:
I see: How to test it?

ndw:
Apply the PR, then run `gradle --no-daemon clean allspecs`. It should run 140 tasks and populate `build/dist` with all the specs.

xml-project:
I never build my local system but always use Travis on my repo. So any change should not affect me. Right?

eriksiegel:
Right. How can I get the changes from this pull request on my machine so I can test them? I'm looking in the help but this isn't mentioned or I overlook it. Anybody with enough GIT knowledge knows?

eriksiegel:
Never mind, i think I found the right help page, finally

ndw:
This works for me, it works for Travis, and I think it's an improvement. Worst case, Erik, it still won't work for you, but if that's the case, let me know.
I'm going to merge it to unblock some other work I'm doing.

eriksiegel:
@ndw It works now on my machine but the command you specified, gradle --no-daemon clean allspecs, took 4:30 minutes! Is that normal?

ndw:
Apparently :disappointed: 
On my Linux box it takes about 30 seconds, but when I ran it on my Windows VM, it took just over four minutes. I assumed that was because my VM has insufficient memory or something, but I guess it's just a Windows issue.
We can set you up with Travis so the builds will happen automatically when you commit to your local repo.

PR #417, opened 2018-06-23
Making XPath 3.1 default

### LINKED ISSUE #408: The xpath-version attribute is no longer required
Right?

xml-project:
Attempt to fix #408 by making XPath 3.1 default


ndw:
I'm ok with this, but I'd also be ok with leaving the attribute as is and saying that in the absence of a value, the value "3.1" is to be assumed.

ndw:
The problem with default attribute values is that they're only present if the document is validated. And in the case of RELAX NG, only if it's validated using RELAX NG DTD Compatibility. I think my preference would be to say it in the prose and leave the grammar unchanged, but I don't object to changing the grammar. It has to be said in the prose as well, though, since I don't think we're going to mandate implementations to do grammar validation of pipelines.

xml-project:
OK, will change this.
Since I didn't get your argument, may be you could explain it to me in Thursdays telcon.

PR #422, opened 2018-06-23
Attempt to finesse TVT nodes in non-XML content

### LINKED ISSUE #398: Finesse the discussion of creating nodes with TVTs
In 2.10.2 Text Value Templates, we say that TVTs can produce nodes. Make sure that the prose is coherent when the content type of the `p:inline` is, for example, "application/json".

ndw:
Attempt to close #398 but consider the question asked in my comment on that issue!

### DIFF
--- a/xproc/src/main/xml/references.xml
+++ b/xproc/src/main/xml/references.xml
@@ -19,6 +19,7 @@
     <bibliomixed xml:id="xml-serialization"/>
     <bibliomixed xml:id="xpath31-functions"/>
     <bibliomixed xml:id="xslt20"/>
+    <bibliomixed xml:id="xquery10"/>
     <bibliomixed xml:id="xmlschema-1"/>
     <bibliomixed xml:id="xmlschema-2"/>
     <bibliomixed xml:id="xml-id"/>
--- a/xproc/src/main/xml/specification.xml
+++ b/xproc/src/main/xml/specification.xml
@@ -5303,24 +5303,11 @@ the <tag class="attribute">document-properties</tag> attribute
 can be used to set the <glossterm>document properties</glossterm> of
 the provided document.</para>
 
-<para>The document's content type is determined statically in the
-following way:</para>
-
-<orderedlist>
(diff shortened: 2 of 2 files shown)


ndw:
This PR needs to be reworked in light of #426. Or maybe discarded in favor of a different one.

xml-project:
Hi gentleman,

looks good to me. The only thing that I am NOT perfectly comfortable with is:

"(2) From the value of the content-type key in the document-properties map, if the value is known statically."

What does is mean, that the value is known statically?

- document-properties="map{'content-type': 'application/json'}" is fine, but what about
- document-properties="map{'key'  : $a , 'content-type': 'application/json'}" (requires sub-parsing of the map)

- Third
<p:variable name="key" select="'content-type'" />
<p:variable name="value" select="'application-json'" />

<p:input port="source">
	<p:inline document-properties="map{$key:$value}">


I some sense the content type is known statically, but it would require to EVALUATE prolog variables
during static analysis.

To reduce vagueness in the specs and complexity of implementation I would argue to remove 
clause (2) and say it is either "application/xml" or specified by "content-type". 
During dynamic evaluation XD0062 is raised if necessary.

ndw:
I'm happy with that, @xml-project .

ndw:
The content of a TVT has to be an XPath expression and XPath doesn’t directly support node constructors, so I don’t think the problem you outline can arise. You can’t write

    <p:variable name=“foocount” select=“count(<foo/>)”/>

and you can’t write

    <p:inline>
        <foocount>{ count(<foo/>) }</foocount>
    </p:inline>

(You can’t write `{ some text <node /> }` because it isn’t even a vaguely valid XPath expression.)

ndw:
Also close #425.

xml-project:
@ndw Is there a reason you do not merge pr #422 ?

ndw:
I wasn't sure we had consensus on the latest, but I'm game. We can always change it with future PRs.

PR #430, opened 2018-06-25
Added static error for p:empty combined with other bindings

ndw:


ndw:
Thanks. I'll update the test on my next push.

Issue #433, opened 2018-07-02
Remove globbing on p:load and p:document, add rudimentary p:load-directory-list


eriksiegel:
Partial implementation of #325

eriksiegel:
This issue is in retrospect unnecessary, because I handled all stuff mentioned in #325

eriksiegel:
COntinuing in #325

Issue #325, opened 2018-03-11
Problems in URI globbing


xml-project:
While I like the idea of having globs in XProc 3.0, I have to admit, that I do not think our current approach is right:
1. Although we use the term URI globbing, a string like "file:///users/*[0-9].*" is no valid URI ('[' and ']' are not allowed in a URI).
1. Using "?" as the glob mark for any single character conflicts with "?" as character indicating the query string of a URI: Is "http://www.xml-project.com/?0" a URI glob or a URI with a query string?
1. Implementing globs in the way currently defined is very inefficient because almost no (Java) standard library can be used (URISyntaxException) and so for a lot of standard things new code has to be written.

I would therefor like to propose another approach to globbing which is clearly inspirited by ANT:
1. Attributes "href" (on p:load, p:document and c:request) or "path" (on p:directory-list) are proper URIs.
1. Each of these elements get a new optional parameter which contains a globbing-expression. This expression is a whitespace separated list of globs which are used to match children etc. of the directory (file) or collection (webdav) named by the URI. 
1. A child is included in the result, if it matches any of the globs.

I hope, I was able to sketch my idea. I not, please feel free to ask.

eriksiegel:
I agree with Achim that allowing globs in @href would be a step too far. It would IMHO deviate too much from what is the usual contents of a @href and therefore it would be both hard to explain and error-prone to use.

A seperate @glob (or something like that) would solve that problem. We could also go for the ant <fileset> approach (with child elements).

If we go for @glob, would *space* separated be ok? A space is a valid character in a filename...

xml-project:
@eriksiegel Yes, space separation may be wrong because you would have to use %20 to escape spaces in filenames.
The other option is an XDM sequence (xs:string*), so one could write
glob="('*.xml', '*.xsl', '*.xpl')"

ndw:
If we're going to take it out of the URI string, then I don't think it should be called a glob anymore, nor is there any compelling reason for it to follow the globbing rules. It would more easily be described and implemented as a regular expression over the URI string(s). But it introduces additional problems.

First, "/d/**/*.txt" matches "/d/" followed by any number of path segments followed by any filename that ends in ".txt". How can that be expressed as a literal URI and an additional set of match strings? How is the base URI going to be construed as something which matches more than one thing? I guess it could
be the root: `file:///d/` and the match `**/*.txt`. Maybe.

[...]

PR #434, opened 2018-07-02
Removed globbing on p:load, added rudimentary p:load-directory-list. …

eriksiegel:
…See #433 and #325

### DIFF
--- a/steps/src/main/xml/specification.xml
+++ b/steps/src/main/xml/specification.xml
@@ -82,6 +82,7 @@ steps is assumed; for background details, see
 <xi:include href="steps/insert.xml"/>
 <xi:include href="steps/label-elements.xml"/>
 <xi:include href="steps/load.xml"/>
+<xi:include href="steps/load-directory-list.xml"/>
 <xi:include href="steps/make-absolute-uris.xml"/>
 <xi:include href="steps/namespace-rename.xml"/>
 <xi:include href="steps/pack.xml"/>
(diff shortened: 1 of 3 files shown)


xml-project:
... and I would suggest to have the regex with as="xs:string+", so I can write more then one regex. Semantics: Something is included, if it matches one of the include-filters and NONE of the exclude-filters.

eriksiegel:
@xml-project This is how I remember it. We removed globbing fro p:document and p:load and decided on a special step p:load-directory list that would support globbing. That's also my interpretation of the consensus on #325.

And anyway, the description of p:load-directory-list is almost a verbatim copy of p:directory-list. So your comments on include/exclude filters etc. apply also on p:directory-list.

xml-project:
@eriksiegel I will never ever question your memory. I am just not sure what "Do away with globbing" in the minutes means.

Concerning p:directory-list and include/exclude-filter: You are certainly right, and I remember me having made this suggestion, but I should record it as an issue.

xml-project:
I just read Norm's addition to the minutes and I still do not have a clue, what "Do away with globbing" means.
Proposal: Lets merge the pull request and talk about it in next telcon.

eriksiegel:
@xml-project Good suggestion. Unless anybody objects, I will merge the pull-request somewhere tomorrow

gimsieke:
Re https://github.com/xproc/3.0-specification/pull/434#issuecomment-402022338: I think “do away with globbing” was only aimed at `p:load` and `p:document`. Another possible explanation is that it says “use regexes instead of glob patterns”. It can well be that i wrote this sentence first, when there was consensus not to glob `p:load` and `p:document`, and then I didn’t update it in the light of the more nuanced discussion that ensued.

PR #442, opened 2018-07-09
Attempt for fix #439

### LINKED ISSUE #439: Non empty text nodes as siblings of implicit inlines
Currently the [specs](http://spec.xproc.org/master/head/xproc/#err.S0079) says:

> It is a static error if comments or processing instructions occur as siblings of an element node that would be treated as an implicit inline.

According to the RNG-syntax text nodes are not allowed either, but we do not have an error for this.

Proposal: Change XS0079 to:

 It is a static error if comments, **non whitespace text nodes** or processing instructions occur as siblings of an element node that would be treated as an implicit inline.

xml-project:
Attempt to fix #439 to include non-whitespace text nodes as error condition


xml-project:
Are you really serious with that?

xml-project:
Changed as requested!

PR #443, opened 2018-07-09
Attempt to fix #440

### LINKED ISSUE #440: Clarify p:documentation/p:pipeinfo as siblings of implicit inline documents
From a [discussion about the test suite](https://github.com/xproc/3.0-test-suite/issues/20#issuecomment-403310476):

What about `p:documentation` (or `p:pipeinfo`) in this example?
````
<p:identity>
  <p:with-input> 
    <p:documentation>Some documentation.</p:documentation>
    <doc />
  </p:with-input>
</p:identity>
````
@ndw said:

> I propose that we say explicitly that they are not treated as inline content; they are ignored.
>If you want an inline p:documentation or p:pipeinfo element, you have to put it inside an explicit p:inline wrapper.

xml-project:
Fixing #440 by explicitly saying that p:documentation and p:pipeinfo are ignored.


xml-project:
Yes, that is better.

xml-project:
Changed as requested.

Issue #448, opened 2018-07-12
Static error for pipe=""


ndw:
The spec says, of @`pipe` on `p:with-input`:

> The attribute value *must* be space-separated list of port@step pairs of the form “port-name@step-name”.

Which is a lie because either port-name or step-name may be omitted. I propose:

1. Clarify the initial sentence so that it isn't a lie.
2. Add a static error for the case where the form of the pipe attribute is not a list of names, in particular, I think `pipe=""` should be an error.

ndw:
In fact, you know what, I'll just make the PR and we can argue about it there. :smile:

ndw:
Nevermind. I'm still going to draft an editorial improvement, but I see that pipe="" is given an explicit semantic later on in the prose.

PR #449, opened 2018-07-12
Clarify pipe and href attributes; add XS0090

ndw:
I think the only possibly controversial part of this proposal is that I describe @href and @pipe under `p:with-input` and then everywhere else, I say "as described in p:with-input`.

I think the descriptions are sufficiently complicated that I'd prefer to avoid repeating them. But I'm prepared to be out-voted.

### DIFF
--- a/xproc/src/main/xml/specification.xml
+++ b/xproc/src/main/xml/specification.xml
@@ -4103,35 +4103,48 @@ empty sequence.</error></para>
 </varlistentry>
 <varlistentry><term><tag class="attribute">href</tag></term>
 <listitem>
-<para>The <tag class="attribute">href</tag> attribute is a shortcut for a
-<tag>p:document</tag> child element with the specified
-<tag class="attribute">href</tag> attribute. <error code="S0081">If
-<tag class="attribute">href</tag> is specified, it is a <glossterm>static error</glossterm>
-if <tag>p:with-input</tag> has any child elements other than <tag>p:documentation</tag> and
-<tag>p:pipeinfo</tag>.</error></para>
-<para><error code="S0085">It is a <glossterm>static error</glossterm> if a <tag>p:with-input</tag>
-has a <tag class="attribute">href</tag> attribute and a <tag class="attribute">pipe</tag>
-attribute.</error></para>
+<para>The <tag class="attribute">href</tag> attribute is a shortcut for one
+or more <tag>p:document</tag> children. The attribute value is a whitespace-separated list
+of URIs. Each URI is a shortcut for a <tag>p:document</tag> child element with the
+<tag class="attribute">href</tag> attribute containing the URI. If more than one URI
+is present, the URIs are processed left-to-right and the effective sequence of
+<tag>p:document</tag> children preserves this order. If the value is empty, it is
+treated as a single, empty URI.</para>
+
+<para><error code="S0081">If <tag class="attribute">href</tag> is specified,
+it is a <glossterm>static error</glossterm> if
+any child elements other than <tag>p:documentation</tag> and
+<tag>p:pipeinfo</tag> are present.</error></para>
+
+<para><error code="S0085">It is a <glossterm>static error</glossterm> if both
+a <tag class="attribute">href</tag> attribute and a
+<tag class="attribute">pipe</tag> attribute are present.</error></para>
 </listitem>
 </varlistentry>
 <varlistentry><term><tag class="attribute">pipe</tag></term>
 <listitem>
 <para>The <tag class="attribute">pipe</tag> attribute is a shortcut for one or
 more <tag>p:pipe</tag> children. The attribute value <rfc2119>must</rfc2119> be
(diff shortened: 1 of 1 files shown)


ndw:
Related to #448

ndw:
I think we decided that in the past and just forgot to write it down.

xml-project:
As href is marked as AVT and we say its a space separated list and the result of an avt is an xs:untyped-atomic: What is the precise semantics? 

The result of evaluating the avt is taken as a string, split into chunks by using ' ' and then every chunk is turned into an xs:anyURI. Right?

This means I not statically determine how many documents will be loaded by a @href. Right?

ndw:
I propose that if it contains AVTs, each template is a "uri". So

     href="a.xml {$foo} b.xml"

is equivalent to:

     <p:document href="a.xml"/>
     <p:document href="{$foo}"/>
     <p:document href="b.xml"/>

Because otherwise the semantics are not at all like a sequence of `p:document` children, AFAICT.

xml-project:
But then it is not a space separated list of uris, but a space separated list of AVT resulting in uris

Consider "a.xml {concat ('b', '.xml')}"

If I do space separation first, this will be "a.xml", "{concat " ...

xml-project:
So :`href="a.xml {$foo} b.xml"` with $foo="x.xml y.xml" is an error, because it will result in
````
<p:document href="a.xml" />
<p:document href="x.xml y.xml" />
<p:document href="b.xml" />
````
where the second p:document has no valid uri

**Comment by ndw** [comment]
That's the consequence of my proposal. Actually, I think we'll piss users off if we do that and the second document should be the URI "x.xml%20y.xml".

If we want to allow space separated values in AVT expressions, we could say that p:document takes a space separated list.

**Comment by gimsieke** [comment]
If URIs are not allowed to contain spaces (which I think is true, as opposed to file system names), we can just say that each token, after tokenization, is treated as a URI and must be castable as xs:anyURI.

**Comment by xml-project** [comment]
@gimsieke How is this different from my proposal?

**Comment by xml-project** [comment]
@ndw I like your approach and could live with the consequences, but as AVT can contain spaces, it is not a space separated list of uri's.

**Comment by ndw** [comment]
I thought I finessed the space-separated-list of URIs thus: 

> The attribute value is a whitespace-separated list of URIs (if the value includes attribute value templates, each template is taken as a single “URI”). 

Is there a consensus developing here that the URIs in p:document's @href should be allowed to be a space separated list as well (post AVT expansion, presumably).

**Comment by ndw** [comment]
Okay. I'll look into that.

**Comment by ndw** [comment]
Right. Please look at the last changes and give me another explicit thumbs up to merge. Or not, you know. :smile:

**Comment by eriksiegel** [comment]
I really don't understand why we need additional semantics around AVTs other than just string expansion.... It confuses things and makes it harder to explain. I think I'm missing something here.

**Comment by xml-project** [comment]
Sorry, I lost the overview over the different changes. @ndw can you provide a link to a formatted spec with all changes in this PR. I didn't manage for find one in your public ghpages.

**Comment by ndw** [comment]
@xml-project http://xpspectest.nwalsh.com/xs0090/head/xproc/

**Comment by ndw** [comment]
I don’t understand where I’m failing to communicate. The `href` attribute on `p:document` in my proposal does just what you suggest. It expands AVTS then treats the value as a whitespace separated list of URIs.

In the status quo draft today, this is an error:

```xml
<p:variable name="foo" select="'a.xml b.xml'”/>
…
  <p:document href="{$foo}"/>
```

unless you happen to have a document named “`a.xml b.xml`” lying around. My proposal fixes that by saying `p:document` parses `href` after AVT expansion.

The other part of my proposal is to keep the semantics of `href` as syntactic sugar for something else. That means something that gets rewritten statically. In my case *literally* rewritten statically.

If we can’t make that work, then we can’t call it syntactic sugar and we have to do something else.

xml-project:
@ndw Sorry, completely my fault (Please remember I am not a native speaker, so most of the time, _I_ am failing to communicate): 

What I wanted to say is: I agree with your proposal on p:document. 

I failed to see, why we need another solution on attribute "href". Now I got it, because you explained it in your comment to Erik.  (Read this _after_ I wrote my comment.)

Proposal:

1. Extend your proposed changes to href on p:input (I think this is an oversight).
2. Merge it!
3. Let's think about adding a note to attribute "href" explaining, why AVTs are handled slightly different here.

ndw:
Right. I've fixed the oversight on `p:input`, thanks Achim, and if it passes tests, I'll merge it unless I hear an objection from anyone.

eriksiegel:
And if this is all caused because we mention it a "syntactic shortcut" then drop the term and the semantics.

@href gets its AVTs string-expanded, tokenized on whitespace and the result is treated as a sequence of URIs

ndw:
Right. Now that we allow a list in @href on p:document, the problem solves itself by moving the whole href there.

PR #455, opened 2018-07-15
First attempt at static variables and options

### LINKED ISSUE #452: Re-instate static variables as solution for the "variable in prolog" problems?
I was going over the use cases I had for wanting variables in the prolog. And most of them (if not all) were relatively simple cases of wanting to declare/define things only once and/or use a descriptive name for something. Computation using run-time-available-only values (like options) are, for me, *not* in scope.

So my thoughts went to: Why don't we re-instate the idea of static variables (like XSLT has, see also https://github.com/xproc/3.0-specification/issues/138) with the following semantics:

- Static variables *must* (of course) be compile-time computable (whether an implementation really does this at compile time or by lazy evaluation at run-time is of course implementation defined). So no references to options, ports, etc. Only to other static variables.
- Variables in a step prolog and as global library variables are allowed to be static *only*. You *must* specify `static="true"` (or whatever syntax). 
- Because of the static nature of the variables we might (not sure) redo the content model of a p:declare-step back to "mix everything"?

Yes, this a p:constant in disguise. But I don't mind, it solves what I want these prolog variables for. As a nice side effect it makes the `@use-when`  much more useful. And it sounds more easy to implement???

Very interested whether this would solve @gimsieke 's use-cases.

ndw:
Here's my first try. I think the diffs are readable.

Inspired by the XSLT spec, I've added the notion of static options as well. These are options who's value can be provided but must be provided "at compile time". So you can specify the serialization method as an option, as long as it's declared static.

If we accept this, I think it will close #452.

However, I fully expect there to be questions, so fire away.

### DIFF
--- a/src/main/schema/core30.rnc
+++ b/src/main/schema/core30.rnc
@@ -69,6 +69,7 @@ content-types.attr = attribute content-types { ContentTypes }
 collection.attr = attribute collection { xsd:boolean }
 pipe.attr = attribute pipe { text }
 expand-text.attr= attribute expand-text { xsd:boolean }
+static.attr = attribute static { xsd:boolean }
 
 exclude-inline-prefixes.attr =
    attribute exclude-inline-prefixes { ExcludeInlinePrefixes }
@@ -145,7 +146,8 @@ Library =
       decl.attributes,
       common.attributes,
       use-when.attr?,
-      (Import|DeclareStep|Documentation|PipeInfo)*
+      (Import|ImportFunctions|Documentation|PipeInfo)*,
+      (DeclareStep, (Documentation|PipeInfo)*)*
    }
 
 [
@@ -153,6 +155,17 @@ Library =
 ]
 Import =
    element import {
+      common.attributes,
+      use-when.attr?,
+      href.attr,
+      (Documentation|PipeInfo)*
+   }
+
+[
+   sa:class = "language-construct"
+]
+ImportFunctions =
+   element import-functions {
       common.attributes,
       use-when.attr?,
       href.attr,
@@ -319,7 +332,22 @@ OptionSelect =
    element option {
       name.qname.attr,
       as.attr?,
-      required.attr?,
+      static.attr?,
+      attribute required { "false" }?,
+      select.attr,
+      common.attributes,
+      use-when.attr?,
+      (Documentation|PipeInfo)*
+   }
+
+[
+   sa:class = "language-construct"
+]
+StaticOption =
+   element option {
+      name.qname.attr,
+      as.attr?,
+      attribute static { "true" },
       select.attr,
       common.attributes,
       use-when.attr?,
@@ -363,16 +391,17 @@ WithOption = WithOptionSelect
 [
    sa:class = "language-construct"
 ]
-VariableSelect =
+GeneralVariable =
    element variable {
       name.qname.attr,
       as.attr?,
+      static.attr?,
       select.attr,
       [ sa:avt = "true" ]
          collection.attr?,
       [ sa:avt = "true" ]
          href.attr?,
-         pipe.attr?,
+      pipe.attr?,
       common.attributes,
       expand-text.attr?,
       inline.attributes,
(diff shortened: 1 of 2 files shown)


eriksiegel:
+ 1 from me. 

However I have a gap in my understanding: Provide values for static options at compile time... How?

ndw:
How those values are provided at compile time is implementation dependent.

As of about 10 minutes ago, since my implementation compiles and runs in one step, you just put them on the command line and I do "the right thing".

gimsieke:
Maybe I now have a gap in understanding what static options are. 
My understanding until [your answer](https://github.com/xproc/3.0-specification/pull/455#issuecomment-405219288) was that I can supply statically evaluable expressions as static options to a step when I invoke this step.
Example: In `<foo:bar baz="{'goo'}"/>` `*` when `baz` is declared a static option of `foo:bar` (with a default value of, say, `baz="'loo'"`), the processor will accept the value `goo` because it doesn’t refer to any document or other non-static options/variables. Is this how static options can be overridden? 
What does the processor do if someone tries to invoke `<foo:bar baz="{/*}"/>`? Is it supposed to throw an error (I’d expect something like „XD0026: Expression refers to context when none is available: /*“, but it needs to throw a static error)?
Similarly, if there is `<p:option name="zoo" static="false"/>` in the prolog of a pipeline that invokes `foo:bar`, an attempt to refer to it in its subpipeline, `<foo:bar baz="{$zoo}"/>`, might lead to „XD0023: Undeclared variable in XPath expression: $zoo“, but this again is a dynamic error, not a static one that will be reported when the processor analyzes the pipeline, before trying to execute it.
So I now think that I didn’t get exactly what you were proposing as static options and that I don’t understand the remark about implementation-dependence.
Among other things, a processor must make sure _statically_ that only non-static options and variables will be part of the [context](http://spec.xproc.org/master/head/xproc/#xproc-xpath-context-31) and that these options may even be evaluated when there is no document in context. These requirements don’t look implementation-dependent. 
Is any processor able to perform these checks statically, without trying to execute the pipeline? Does XProc 3.0 stipulate that any conforming processor must be able to perform this analysis?
If static options are supposed to work like I assumed they would, we need to have two static errors like XD0023 and XD0026. In particular, we should say in the [context](http://spec.xproc.org/master/head/xproc/#xproc-xpath-context-31) section that when overriding the default value of a step’s static option during invocation, non-static options and variables will be excluded from the context’s in-scope variable bindings.

`*` I could have written, equivalently, `<foo:bar baz="goo"/>` or `<foo:bar><p.with-option name="baz" select="'goo'"/></foo:bar>`

ndw:
Think of a processor as having two phases: a static phase and a dynamic phase. You can provide values for static options during the static phase. Consider:

```xml
<p:declare-step name="something" type="x:bar">
  <p:option name="method" select="'xml'" static="true"/>
  <p:output port="result" serialization="map { 'method': $method }"/>
  ...
</p:declare-step>
```
If you provide a different value for `$method` at static analysis time (mechanism implementation defined), then you can override that value and make it something else.

You can never change the option value dynamically:

```xml
  <x:bar method="xml"/>
```

will always be an error.

xml-project:
@ndw I am still trying to understand the concept of static options/variables you introduced.
The example and your comments are very helpful for this. 
May I, without any attempt to criticize any aspect of your proposal, but in order to understand what is proposed, aks two questions:

(1) What happens (in your example), if I turn `static` for 'method' to false? (Apart from losing the ability to overwrite the value statically.) Will this be an error because the XPath-content for `serialization` only contains static options and variables?

(2) Since `p:variable static='true'` can have a p:document-binding: Does this mean the document is loaded during static analysis?

gimsieke:
It was not obvious to me what happens during static analysis. I also didn’t know that when using the equivalent of `p:with-option`, even if its `@select` XPath expression contains only statically known bindings or expressions that are legal in `@use-when`, and even if does not refer to any documents, is always considered “dynamic”. Is it just me or do we need to explain “static analysis” and “can be evaluated statically” better?

ndw:
@xml-project : 
1. If you remove static in the example above, you'll get err:XS0091 because a non-static option precedes `p:output`.
2. Yes.

@gimsieke : 
That's not your fault, we need to write that section of the spec.

ndw:
See #458

xml-project:
@ndw OK, I see the concept. 
XS0091 says: 

> It is a static error (err:XS0091) if a p:variable or p:option that occurs before the subpipeline is not declared static.

So p:option (for non statics) is now part of our the subpipeline?

ndw:
Uh, I may have messed that up.
I was trying to avoid introducing new vocabulary.
No, I don't think they are part of the subpipeline, they just have to come after input/output.
And the reason I was trying to forbid them is to avoid complicated and non-intuitive rules about scoping. I'll file a bug.

xml-project:
Ok, I see. 
This was the point I am struggling with to see the relevance of static options. So what propose as a content model for p:declare-step is something like:
````
(p:import | 
     p:import-functions)*,
    (p:input | 
     p:output | 
     p:option | 
     p:variable)*,
     p:option*,
    p:declare-step*,
````
where the second p:option is static="false". 

Yes, I think you found a very elegant solution!

ndw:
I think it covers the use cases really well without causing implementors to weep. But the prose for this is going to be a bit tricky to write. See #459 as well.

xml-project:
`p:static-option` might save a lot of problems with the prose.

ndw:
:smile:

PR #466, opened 2018-07-20
Make href on p:document a single URI

### LINKED ISSUE #453: Revert multiple URIs in @href on p:document?
Multiple URIs on p:document was a workaround for dynamic variables in @href on p:input.

Now that we don't have that, I propose that we make the @href attribute on p:document a single URI again.

That will avoid the complication of multiple URIs which have to share the same properties.

ndw:
I also made static analysis and dynamic evaluation into termdefs.

Close #453


ndw:
No. Good catch.
Total brain cramp on my part. I've confused the general case of p:document with p:document in a default pipeline input. Nevermind. Will fix.

ndw:
Replaced by #495

PR #485, opened 2018-08-11
Attempt to fix-484

xml-project:
Introducing new static error (S0097) raised, when @select on p:option is syntactically invalid or makes reference to context node, size, or position.


xml-project:
Overtaken by evens (issue is already closed)

PR #494, opened 2018-08-11
 Clarify p:cast-content-type wrt JSON and HTML

ndw:
Replaces #465


ndw:
@xml-project I've clarified the HTML and JSON cases.

With respect to parameters, I think the point of having them is so that implementations can be told how to do the conversions. I don't think we need to have a special key with a value for that purpose. I think the absence of any parameters can have that purpose.

I did try to clarify that the conversion functions in XPath 3.1 *should* be used where appropriate.

ndw:
Ok. Can you start a wiki page with a Leipzig agenda? I’m on my phone.

xml-project:
@ndw Done!

PR #498, opened 2018-08-12
Attempt to fix #450

### LINKED ISSUE #450: XS0037 vs. xs0044
Currently XS0037 says:

> It is a static error (err:XS0037) if any step directly contains text nodes that do not consist entirely of whitespace.

Is there a reason for the restriction to _step directly_? Consider
````
<p:identity>
  whitespace1
  <p:input port="source">whitespace2 
````
'whitespace1' raises XS0037, while 'whitespace2' has to be XS0044 (in lack of a bettet one).

I think this makes lexing difficult and adds no value to authors.

xml-project:
Fixing #450 by changing the error text of XS0037 to cover all cases of non-whitespace text nodes.


ndw:
:+1:

Issue #486, opened 2018-08-11
@as and @select on p:option


xml-project:
Is it a static error if the evaluation of `@select` on p:option return a value that does not match the sequence-type in `@as`?

Option 1: Yes, we need to introduce a new static error for this.
Option 2: No: We wait until the default value of the option is used and raise XD0036 then.

gimsieke:
To check the sequence type during static analysis would certainly be more anal, as implied by “analysis.” 
Jokes aside. Since only the `@select` attributes of static options and variables need to be evaluated during static analysis, I’d say we cannot require a processor to evaluate _all_ options statically in order to determine the return type for the default values. For static variables and options, I think a sequence type error should be raised during static analysis though.

xml-project:
>  I’d say we cannot require a processor to evaluate all options statically in order to determine the return type for the default values.

Contrary to your reading, I would say the processor has not just to determine the static type of `@select` but it's value:

> The select expression is evaluated during static analysis.

So the processor knows for all options (static or not) whether the default value matches the sequence type. The only question IMHO is whether this knowledge is used for a static error or not.

What did I miss?

gimsieke:
In http://spec.xproc.org/master/head/xproc/#p.option it says indeed “The select expression is evaluated during static analysis.” It seems to apply to all options. 

On the other hand, there is http://spec.xproc.org/master/head/xproc/#statics “The values of static options and variables are computed during static analysis.” and http://spec.xproc.org/master/head/xproc/#static-expressions “Several kinds of expressions are evaluated during static analysis: … The select expressions on static p:options, unless a static default value has been provided.”
I think if all select expressions need to be evaluated anyway during static analysis, it need not be mentioned that the select expressions of static options need to be evaluated during static analysis.

Another indication the the rule “evaluate select during static analysis“ is not intended to be this general: The first sentence of the para that immediately precedes “The select expression is evaluated during static analysis.” reads “A p:option with a may be declared “static”.” Maybe “with a” is superfluous here? I can imagine that if the editing has been sloppy in the preceding para, then the sentence about evaluation of all options’ select expressions during static analysis is not intended to be this general, but it should only apply to static options. This assumption is corroborated by the fact that it was added in the same [commit](https://github.com/xproc/3.0-specification/blame/master/xproc/src/main/xml/specification.xml#L4899) as the static option stuff. Norm will be able to clarify this.

xml-project:
[...]

PR #499, opened 2018-08-12
Attempt to resolve he confusions in #486

ndw:
I've tried to clarify the static variables and options.

In order to simplify the list in 7.1, I've made the content-type of p:document explicit (parallel to p:inline). This avoids having to make evaluation of document-properties static, which I don't think makes any sense.

### DIFF
--- a/xproc/src/main/xml/specification.xml
+++ b/xproc/src/main/xml/specification.xml
@@ -4815,7 +4815,7 @@ node for the expression that defines the value of the variable.</error>
 <section xml:id="p.option">
 <title>p:option</title>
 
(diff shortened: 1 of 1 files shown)


ndw:
It may be useful to look at the diff version: http://xpspectest.nwalsh.com/iss-486/head/xproc/diff.html

ndw:
@gimsieke I fixed the obvious typos. I added the question about options in libraries to the Leipzig agenda. I'm leaving the "common attributes" paragraph alone for the moment; it's used in several places, we should try to fix it globally if we don't like it.

PR #500, opened 2018-08-12
Remove leftover references to character maps

### LINKED ISSUE #492: Leftover from p:character-maps
On [Serialization parameters](http://spec.xproc.org/master/head/xproc/#serialization) we still have a reference to p:character-map in "use-character-maps".

We agreed to use the map{} as XPath does for [fn:serialize()](https://www.w3.org/TR/xpath-functions-31/#func-serialize), if I remember right.

ndw:
Close #492

### DIFF
--- a/xproc/src/main/xml/specification.xml
+++ b/xproc/src/main/xml/specification.xml
@@ -2230,9 +2230,6 @@ evaluating the template.</para>
 <para><option>undeclare-prefixes</option>: false</para>
 </listitem>
 <listitem>
-<para><option>use-character-maps</option>: ()</para>
-</listitem>
-<listitem>
 <para><option>version</option>: 1.0</para>
 </listitem>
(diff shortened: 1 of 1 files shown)


ndw:
@xml-project I'm confused. Where? I don't see any references to character maps in the `iss-492-charmaps` branch: http://xpspectest.nwalsh.com/iss-492-charmaps/head/

(I'm probably being blind, but ...)

xml-project:
@ndw Sorry for being not precise: The problem is that you deleted the whole section on the serialization "use-character-maps". I think we still have this parameter (key in the map), but it is not connected to "p:character-map" anymore, but to a "map(xs:string, xs:string)?"
Is this a better explanation of the problem?

ndw:
Right. I just got this entirely wrong. Closing this PR and trying again.

PR #510, opened 2018-08-20
Attempt to resolve p:inline issues

### LINKED ISSUE #508: @expand-text of descendants of (implicit) inlines
Currently we [say](http://spec.xproc.org/master/head/xproc/#expand-text-attribute):

> If a p:inline occurs as a descendant of another p:inline, any expand-text within the inner p:inline is treated like an ordinary attribute; that is, neither will its value influence text expansion nor will it be removed from the descendant.

Since we have now `@expand-text` everywhere, I am not sure this special treatment of `p:inline` is a legacy (and should be expanded to all occurrences of `@expand-text`) or we keep its special role.

ndw:
1. Close #508, the interpretation of the `[p:]expand-text` attribute does not depend on where it occurs with respect to nested `p:inline` elements.
2. Everytime I've read the `p:inline` section recently, it's felt very clunky. I've tried to improve it. Along the way, I believe that I've identified a redundant error.
 
See http://xpspectest.nwalsh.com/iss-508/head/xproc/


xml-project:
I think (2) is ok, but I am not sure about (1). The current specs say:

> If a p:inline occurs as a descendant of another p:inline, any expand-text within the inner p:inline is treated like an ordinary attribute; that is, neither will its value influence text expansion nor will it be removed from the descendant.

Your suggestion is to completely remove this passage, but I think this would make every `@expand-text` inside an (implicit) `p:inline` active, wouldn't it? 
@gimsieke suggested replacing the section above with:

> Any expand-text attribute on a descendant of p:inline or within an implicit inline is treated like an ordinary attribute; that is, neither will its value influence text expansion nor will it be removed.

> Any p:expand-text attribute, however, if its prefix p is bound to the XProc namespace, on a descendant of p:inline or within an implicit inline will influence expansion and will be removed in the resulting document.

Wouldn't this be a better approach?

ndw:
I could live with that. I figured we'd just treat the `[p:]expand-text` attribute uniformly. Given that `p:inline` occurring as a descendant of `p:inline` is a vanishingly small use case, I don't think we need a special rule for it.

ndw:
Let me make one more attempt to defend the changes in this PR.

Note that in every case below, we’re talking about elements that occur
inside an inline, either descendants of an explicit `p:inline` or
content in an implicit inline.

1. The `p:expand-text` attribute can only occur on elements that are
   not in the XProc namespace. They always controls the expand text
   behavior of its descendants, there’s never anything special about
   them.
2. If an `expand-text` attribute appears on an element that isn’t in
   the XProc namespace, it’s just an attribute. There’s nothing
   special about it.
3. If an `expand-text` attribute appears on an element in the XProc
   namespace (within an inline so this is a real edge case, I expect),
   it controls the expand text behavior; it’s not special.

If a user wants to include literal `p:expand-text` attributes on
elements within an inline; they can name them `q:expand-text` and then
rename them in a subsequent step.

ndw:
P.S. I updated the PR with an example and I added the text that says `[p:]expand-text` are removed. I think that was our understanding, but I don't think the spec said that!

ndw:
Nevermind. We changed our minds about what to do at the 5 September workshop.

Issue #428, opened 2018-06-24
Option "serialization" on p:output for atomic steps


xml-project:
Currently we have:
```
<p:output
  port? = NCName
  sequence? = boolean
  primary? = boolean
  content-types? = ContentTypes
  expand-text? = boolean
  serialization? = map(xs:QName,xs:anyAtomicValue) />
```
for atomic steps, but I wonder what "serialization" is for.

> If the processor is not serializing (if, for example, the pipeline has been called from another pipeline), then the serialization must be ignored.

And since an atomic step is always called from another step, the serialization options are always ignored.

Is there any (non-propertary) use case for this? I could think of calling an atomic step from within a p:library, but this is not a standard feature of a processor.

gimsieke:
I don’t think that your presupposition that an atomic step is always called from another step is justified. At least for Calabash, there’s nothing wrong with calling, for example, `-s p:xslt` on the command line. Then the serialization option, at least on the `result` port, is meaningful.

xml-project:
According to the specs, an atomic step is ALWAYS called from another step. 
What you mentioned is what I called a proprietary use case in my comment. In MorganaXProc you can call an atomic step from a library, but there is no place in the specs saying either use case is required.

The point is what to answer if someone asks what "serialization" is for and you can not find an answer in the specs. Possible solutions:
(1) Say that the processor may execute atomic steps by a processor specific mechanism.
(2) Remove "serialization". This does not prevent the implementer from doing any magic, but it leaves no un-explained feature in the specs. If you want to set serialization parameters for an output port: Write a pipeline!

xml-project:
@gimsieke Do you prefer to keep it and add a note saying it is used, when processors are able to call atomic steps?

xml-project:
@eriksiegel 

> ButI don't understand: an atomic step is ALWAYS called from another step? There must always be a first step that is "called", "processed" or whatever.

Yes, but this step is a compound step and for compound steps serialization makes sense. But we do not say in any place, that e.g. p:add-attribute may be called outside a compound step.

ndw:
I think that’s a bug. We used to have a p:serialization element to define serialization parameters. When we made it a map, we must have just stuck it on p:output generally.

Atomic steps don’t serialize their output. Even in XML Calabash where you can run a step “directly”, I believe that’s described as syntactic sugar for a p:pipeline containing a single step.

[...]

PR #533, opened 2018-09-06
Attempt to resolve issue 428

ndw:
* Refactor the `p:declare-step` and `p:output` patterns in `core30.rnc` so that there are different versions for compound and atomic step declarations and different versions of the output declaration for atomic steps, compound steps, and declarations for pipelines
* Refactor the prose to reflect these changes.


ndw:
Close #428

Issue #388, opened 2018-05-04
Attributes and options with map value


xml-project:
From a conversation with @gimsieke:
We decided that every attribute value on an XProc element and every option on an XProc step that has a map, array (or function item?) type, is interpreted as an XPath expression evaluated to produce the value.

This decision is currently not reflected in the specs.

ndw:
Note that this applies to the options on user-defined steps if they have a type of map or array!

PR #535, opened 2018-09-07
Text changes for issue #388

eriksiegel:
I tried to describe map attributes and moved the section about option attributes to the right location


eriksiegel:
I now made "map-attribute" an official termdef. Better?

eriksiegel:
Ok, datatype is now sequence type.
Also changed the wording a bit because I saw the definition of AVT was different then I thought it was.

gimsieke:
[Men Without Hats](https://en.wikipedia.org/wiki/The_Safety_Dance), apparently…

eriksiegel:
Aren't we all (in the committee). I haven't seen a hat on our heads in Leipzig :-)

Issue #293, opened 2018-02-14
Introduce new step p:tee


xml-project:
See [minutes, day 2](https://github.com/xproc/Workshop-2018-02/wiki/Raw-Minutes,-7-Feb#ptee)

eriksiegel:
Proposal for a p:tee step:

A p:tee step acts exactly like a p:identity step with the additional functionality of being able to writing the contents "flowing through" to a file.


```
<p:declare-step type="p:tee">
   <p:input port="source" content-types="*/*" sequence="true"/>
   <p:output port="result" sequence="true" content-types="*/*"/>
   <p:option name="href" required="true" as="xs:anyURI"/>        
   <p:option name="serialization" as="map(xs:string,xs:anyAtomicValue)"/>
   <p:option name="enable" as="xs:boolean" required="false" select="true()"/>   
</p:tee>
```

When @enable is false() the step acts exactly as a p:identity step.

When @enable is true() every document flowing through is written to the URI as specified by @href using the same semantics as p:store. Existing content on @href is overwritten if possible (for files).

An implementation in XProc could be written as:
```
<p:identity name="tee-in"/>
<p:if collection="true" test="$enable">
 <p:store>
    <p:with-option name="href" select="$href"/>
    <p:with-option name="serialization" select="$serialization"/>    
 </p:store>
</p:if>
<p:sink/> 
<p:identity>
  <p:with-input pipe="@tee-in"/>
</p:identity>
```

(NOT SURE: Can the above XProc fragment be done differently (shorter maybe?) using ports on the p:if?)

ndw:
I think the `enable` option is a little odd. If it's for debugging, we have other mechanisms for that. If it's not for debugging, why do you want a dynamic enable condition? And, obviously, you can implement this yourself in XProc if you want precisely these semantics.

eriksiegel:
The enable option is there because I use it for... debugging. I usually have a $debug parameter in my stylesheets that turns tee operations on. 
Yes, you're right, implementation in XProc is easily done (and that's exactly what I have done in the 1.0 case). The request for a p:tee was more out of convenience reasons, I think its a nice lightweight debugging tool, useful for more than just me. But if I'm the only one that does things this way, no problem to skip it.

xml-project:
One reason for having p:tee as a step in the standard library (as opposed to implementing it with XProc in a p:library) might be the debug flag itself. We could think about introducing the notion of a XProc processor running in debug-mode which switches steps as p:tee on.
[...]

Issue #534, opened 2018-09-07
Non-required options and @as


xml-project:
In XProc 3.0 we say, that an option that is declared without an explicit `@select` has the empty sequence as default value.
In the step library (almost) all non required options are declared without occurrence modifier, e.g. "xs:string". But `() instance of xs:string` is false, so the processor must raise an error.

Best way to fix this seems to change occurrence modifier in  `@as` in the step library to something like `xs:string?`.

Did I miss something? Better proposals?

gimsieke:
Agree

ndw:
:+1:

ndw:
We can get rid of the `required` altogether then, can't we?

gimsieke:
I think so

eriksiegel:
Rid of the `required`? What do you mean?

gimsieke:
`as="xs:string"` says it is required, `as="xs:string?"` says it’s optional

ndw:
required=true is equivalent to `as="sometype"`; required=false is equivalent to `as="sometype?"`. If we're going to require the `as` to allow `()` on non-required options (which I think we should) then there's no reason to separately specify that an option is required or not.

The one exception, I suppose is untyped options. But those are equivalent required=false and if youw ant required=true, you can say `as="xs:string"` (without a ?).

eriksiegel:
I'm sorry but i disagree. `required="true"` means you *have* to **specify** the option on the step invocation, either as a `with-option` or as an attribute. You cannot not mention it.

This is as how it works in XSLT as well with required parameters for named templates.

eriksiegel:
`required="false"` means you can leave it out and then the default (from the `@select`) kicks in

eriksiegel:
So `as="xs:string?" ` does *not* mean its optional, it only means it accepts the empty sequence as value.

ndw:
Ooookaaaayyyy...but
```xml
<p:option name="foo" required="true" as="xs:string?"/>
```
Means you must specify a value for `foo` but that value can be the empty string.
```xml
<p:option name="bar" required="false" as="xs:string?"/>
```
Means you don't have to specify the value, and if you don't it will be the empty string.

There's not a whole lot of difference between those two in practice.
```xml
<p:option name="baz" required="false" as="xs:string"/>
```
In principle means that you don't have to specify the option, but if you don't it will get the value `()` and that's not allowed so you'll get an error. So it is required, even though required=false. I think that's a serious usability problem.

[...]

PR #540, opened 2018-09-12
Step description for p:tee

eriksiegel:
Covers #293

### DIFF
--- a/steps/src/main/xml/steps/tee.xml
+++ b/steps/src/main/xml/steps/tee.xml
@@ -14,12 +14,9 @@ version of) its input to a URI. The step outputs its input unchanged.</para>
     <p:option name="enable" as="xs:boolean" select="true()"/>   
   </p:declare-step>
 
-<para>The value of the <option>href</option> option
-<rfc2119>must</rfc2119> be an <type>anyURI</type>. If it is relative,
-it is made absolute against the base URI of the element on which it is
-specified (<tag>p:with-option</tag> or <tag>p:store</tag> in the case
-of a syntactic shortcut
-value).</para>
+<para>The value of the <option>href</option> option <rfc2119>must</rfc2119> be an <type>anyURI</type>. If it is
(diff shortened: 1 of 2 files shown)


steps/src/main/xml/steps/tee.xml
```diff
@@ -0,0 +1,43 @@
...
+if the URI scheme is not supported or the step cannot store to the
+specified location.</error></para>
+
+<para>The output of this step is what appears on its input port, unchanged.</para>
+
+<para>The <option>serialization</option> option is provided to control the
+serialization of content when it is stored. Serialization is described in
+<biblioref linkend="xproc30"/>.</para>
+  
+  <para>If the <option>enabled</option> option is false, no attempt to store the document will be done. The step will
```
gimsieke: will be made

steps/src/main/xml/steps/tee.xml
```diff
@@ -0,0 +1,43 @@
...
+    <p:output port="result" sequence="true" content-types="*/*"/>
+    <p:option name="href" required="true" as="xs:anyURI"/>        
+    <p:option name="serialization" as="map(xs:QName,item()*)"/>
+    <p:option name="enable" as="xs:boolean" select="true()"/>   
+  </p:declare-step>
+
+<para>The value of the <option>href</option> option
+<rfc2119>must</rfc2119> be an <type>anyURI</type>. If it is relative,
+it is made absolute against the base URI of the element on which it is
+specified (<tag>p:with-option</tag> or <tag>p:store</tag> in the case
```
gimsieke: `p:store`?

eriksiegel: must be p:tee

steps/src/main/xml/steps/tee.xml
eriksiegel: ok

eriksiegel:
@xml-project You're right about the application of #534 on non-required options, but in this case it only applies to `serialization` . The `enabled` option has an explicit `@select`. I'll change it.

PR #545, opened 2018-09-27
Added p:message to the spec

### LINKED ISSUE #29: Standardize cx:message attribute
For some kind of non-disruptive debugging, the cx:message extension attribute of XMLCalabash would be nice to have in XProc 1.1. Motivation: See https://github.com/ndw/xmlcalabash1/issues/238

With attribute value templates in the attribute value we could even do better!

eriksiegel:
This resolves #29. I did a bit of refactoring in the text's structure to make it more logical.

### DIFF
--- a/steps-intro/src/main/xml/specification.xml
+++ b/steps-intro/src/main/xml/specification.xml
@@ -153,44 +153,51 @@ linkend="rfc2119"/>.</para>
   
 <section xml:id="common-options">
 <title>Common options</title>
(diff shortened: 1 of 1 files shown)


eriksiegel:
Checks fail but I haven't a clue why. Asked @ndw for assistance.

gimsieke:
Maybe the section entitled “p:timeout“ needs an ID?

Issue #315, opened 2018-03-04
Binding for p:with-option and p:variable 


xml-project:
Opposed to XProc 1.0 the current syntax of [p:with-option](http://spec.xproc.org/master/head/xproc/#p.with-option) and [p:variable](http://spec.xproc.org/master/head/xproc/#p.variable) allow multiple documents to appear as context node for the select-expression. This results in a dynamic error (XD0008), but is no syntax error anymore.
Is this intentional or a mistake, so we need to replace "*" with "?" in the syntax?

xml-project:
@ndw @gimsieke @eriksiegel Nobody any idea?

gimsieke:
I vaguely remember that we wanted to use the first document for evaluation if multiple documents are present.

xml-project:
@gimsieke 
Thanks Gerrit, that was the hint I needed: It's a sequence because of the collection attribute. And the error has to be dynamic (and not static anymore) because collection is an AVT.

PR #547, opened 2018-09-29
Describe the context item for value templates

### LINKED ISSUE #482: Describe context node for p:document and p:inline
Per discussion on 9 Aug:

The p:inline and p:document elements establish a context node from the default readable port (if there is one). This context node is used for descendant AVTs and TVTs. It is not a static error if there is no DRP, it simply establishes that there's no binding for the context (and it will be a dynamic error to refer to it).

It is a static error to refer to the context in expressions that appear where no default readable port is possible (for example, on p:declare-step, p:library, p:input, and p:output).

ndw:
Value templates get their context from the DRP. It is a dynamic error if a sequence of more than one items appears. Output ports may be read by value templates.
Close #482

### DIFF
--- a/xproc/src/main/xml/specification.xml
+++ b/xproc/src/main/xml/specification.xml
@@ -685,11 +685,13 @@ content types, that they accept. Input documents must
                </para>
             </listitem>
          </itemizedlist>
-         <para>When an input accepts a sequence of documents, the documents can come from any
-            combination of these locations.</para>
-         <para>In contrast, output ports are connected when they are referenced by another input
-            port, <glossterm baseform="declared outputs">declared output</glossterm> or other
-            expression and may be connected to:</para>
+
+<para>When an input accepts a sequence of documents, the documents can
+come from any combination of these locations.</para>
+
+<para>In contrast, output ports are connected when they are referenced
+by another input port, <glossterm baseform="declared outputs">declared
+output</glossterm> or other expression and may be connected to:</para>
 
 <itemizedlist>
 <listitem>
@@ -698,6 +700,14 @@ content types, that they accept. Input documents must
 <listitem>
   <para>An option assigned with <tag>p:with-option</tag> or a
 <tag>p:variable</tag> in a compound step.</para>
+</listitem>
+<listitem>
+  <para>A <glossterm>value template</glossterm> in an immediately following step.
+This can be an AVT in an
+<link xlink:href="#option-shortcut">option shortcut</link>,
+an AVT on a <tag>p:document</tag> element, or a
+value template in a <tag>p:inline</tag>.
+  </para>
 </listitem>
 <listitem>
   <para>One of the outputs declared on its container. </para>
@@ -2021,8 +2031,8 @@ circumstances, contain embedded expressions enclosed between curly
 brackets. Attributes and text nodes that use (or are permitted to use)
 this mechanism are referred to respectively as <glossterm
 baseform="attribute value template">attribute value
-templates</glossterm> and <glossterm baseform="text value
-template">text value templates.</glossterm>.</para>
+templates</glossterm> (AVTs) and <glossterm baseform="text value
+template">text value templates.</glossterm> (TVTs).</para>
 
 <para><termdef xml:id="dt-value-template">Collectively,
 attribute value templates and text value templates are referred to as
@@ -2055,7 +2065,7 @@ XPath error code.</para>
 
 <para>
 <error code="D0050">It is a <glossterm>dynamic error</glossterm> if the
-XPath expression in an AVT or TVT can not be evaluated.</error>
+XPath expression in a value template can not be evaluated.</error>
 </para>
 
 <para>
@@ -2064,6 +2074,13 @@ expression in an AVT or TVT evaluates to something to other than a sequence
 containing atomic values or nodes.</error> Function, array and map items are
 explicitly excluded here because they do not have a string representation.
 </para>
+
+<para>The context item used for evaluating value templates comes from
+the <glossterm>default readable port</glossterm>.
+<error code="D0065">It is a <glossterm>dynamic error</glossterm>
+if the context item for a value template is a sequence of more than one
+item.</error> If the value template appears in a context where no default
+readable port exists, then the context item is undefined.</para>
 
 <section xml:id="attribute-value-templates">
 <title>Attribute Value Templates</title>


gimsieke:
I asked twice in comments ([1](https://github.com/xproc/3.0-specification/issues/315#issuecomment-370366753), [2](https://github.com/xproc/3.0-specification/issues/482#issuecomment-412195140)) without receiving a satisfactory answer: Didn’t we intend at some stage to only use the first of multiple documents that are in context for evaluating an expression? This would remove the need for [D0065](https://github.com/xproc/3.0-specification/pull/547/files#diff-2467a4504c4dfe38920ed086a8364518R2080).

xml-project:
@gimsieke I was not able to find the consent you are citing. Can you give me a hint, when and/or why we decided this. Does it apply to p:with-option and option-shuts too?

gimsieke:
We didn’t formally decide it but we discussed it at some point. I don’t remember where and when. Therefore the phrasing “I vaguely remember”. 
It would apply to `p:with-option`, `p:variable`, `p:when`, value templates, etc. 
The question is, in the absence of `collection="true"` and when multiple documents are in context, should we always only consider the first. 
Motivation: Quite regularly when I use `cx:message`, I get an error that multiple docs are in the context of the message attribute. The message often only refers to variables, so in these cases I can evaluate `p:with-option` in a `p:empty` context. If there are multiple documents in context, I cannot use an attribute value template in `@message`, I need to revert to the lengthy `p:with-option`/`p:empty`.
We don’t have to discuss it in the comments of a pull request, I can also raise a new issue.

ndw:
I'm pretty strongly opposed to "first value" semantics. That's what XSLT 1.0 used for sequences and we moved away from that in 2.0 because of the subtle bugs it introduced. (Mostly failure to detect errors, I think.)

However, I see your point wrt cx:message and other places where there are AVTs that don't refer to the context.

I wonder if we could say that the context item is defined if-and-only-if a single item appears. Pro: it would remove the error when you don't refer to the context and preserve the error if you do. Con: it's a weird rule.

gimsieke:
I like the proposed rule. It won't seem weird to pipeline authors most of the time. And if it does, we can still toss an RTFS (read the fscking spec) in their direction.

xml-project:
> Con: it's a weird rule.
Yes, indeed! I am not sure whether the proposed rules makes adoption of XProc easier (is this still a goal?) for people coming from other XPath-host languages like XSLT. We should think about a better idea to solve the cx:message problem.

xml-project:
> I wonder if we could say that the context item is defined if-and-only-if a single item appears.

What about having the rule that the DRP (if present) supplies the context item for a value template if-and-only-if the XPath expression refers to a context item. If value template relies on the context item then it is an error if there is a sequence with more than one item.

Pro: Would solve the problem and sounds better to me. Con: Possibly difficult to implement.

eriksiegel:
I'm sorry to come back to a point by @ndw  made earlier, but I think that from a user's perspective it would make perfect sense to have a "first value" rule: When more than on document appears on a DRP, the evaluation of AVTs only takes the first one into account.
IMHO errors when more than item appears will be rather puzzling i'm afraid. 

As an alternative, can't we state that the DRP "output" is *always* a collection of documents (zero, one or more) and you simply have to write your Xpath expressions correctly?

gimsieke:
Reply to Achim’s comment https://github.com/xproc/3.0-specification/pull/547#issuecomment-425909722 above:

In an email exchange about #482 (between @ndw, @xml-project, and @gimsieke, while @eriksiegel was vacationing), Norm already expressed reservations against this level of analysis wrt XPath expressions:

------

Achim Berndzen <achim.berndzen@xml-project.com> writes:
> Yes, I was thinking about that line too. But then, why do we say that
> this p:inline consumes the DRP?
>
> <p:inline expand-true="true"><doc>{$a+$b}</doc></p:inline>
>
> So the next step would be: A p:inline, an implicit inline or the href
> on p:document only consume the DRP if the used XPath expression refers
> to the context item, the position or the length. But then you could
> image cases, where one branch of an "if" refers to the context and the
> other one does not. Does "if (false) . else 4" depend on the context
> item?

I think we want a simple rule. A p:inline or a p:document could
require the context item, therefore they always consume it. I think
it’s too complicated to say that it depends on the value of
@expand-text or the presence of an AVT or anything like that.

------

(Unquoted text above by Norm)

And to another message in the same thread, Norm replied:

------
"Imsieke, Gerrit, le-tex" <gerrit.imsieke@le-tex.de> writes:
> What is unclear about 3.0 is whether
> – an attribute option that may hold an AVT or
> – one that actually contains an AVT
> will also consume the DRP.

I think the rule should be “may hold”. I do not think we should make
static analysis depend on parsing attribute values in search of AVTs.

------

Therefore I’d assume that Norm again would consider a proposal that involves parsing the XPath expression and deciding whether it refers to a context item as too difficult to implement or too complicated.

xml-project:
@gimsieke Thanks for reminding me about this. I know it wouldn't be a popular approach, but it would be a solution.

eriksiegel:
@ndw Sorry, I'm getting a bit lost in this discussion. Why would your second example be an error? 

And: Does all this means that the context item for a `<p:with-option>` is different than for an option-setting expressed as an AVT? 

BTW, this seems something better postponed to the next call (which I unfortunately cant attend).

ndw:
The second example is an error because of err:XD0008 which has existed since XProc 1.0. In XProc 3.0, we're providing a collection attribute to allow a sequence, but it isn't the context in that case.
No, it isn't different (unless collection is specified) because a sequence is an error in either case.
We can postpone it until a call if we can't reach consensus here.

gimsieke:
@ndw, @xml-project, and @gimsieke agreed that we feel comfortable adopting Norm’s proposal (context defined iff there’s exactly one document present for evaluating the expression). 
Given his authority and decade-long experience in XML processing language specification, we just believed Norm’s argument that “first-value” semantics is error-prone and should be avoided. 

Regarding Achim’s proposal to analyze whether an XPath expression refers to a context item: There was consensus that this will put too high a burden on implementers. 
A dynamic error will be thrown when the context is nullified and the expression refers to the context.
However, care `<rfc2119>should</rfc2119>` be taken by implementations to issue a meaningful error message in this situation, that is, if the context is undefined not just because there is _no document_, but because _more than one_ document is present.

@eriksiegel, can you live with this solution?

eriksiegel:
I'm trying to understand the implications of this...

So we have a step for which the DRP is one of:

1. Non-existent
1. Empty (no doc appearing)
1. Multiple docs appearing

In all these cases, if I understand things correctly, any AVT is invalid, even simple ones like `{1 + 2}` and `{ $somevar }`?

And what about `p:with-option/@select`? It has no context item either.

If I'm right (and probably not) it would mean you can hardly do anything meaningful if one of the three conditions is true.

gimsieke:
No, AVTs are valid in these cases as long as they don’t refer to a context item.

eriksiegel:
Ah, I think I mis-interpreted:

> Regarding Achim’s proposal to analyze whether an XPath expression refers to a context item: There was consensus that this will put too high a burden on implementers.

That would refer to static analysis *only*, right?

gimsieke:
Yes. If an XPath expression refers to the context item, a dynamic error will be thrown. I think it will currently be [XD0050](http://spec.xproc.org/master/head/xproc/#err.D0050) for value templates, but maybe there should be other dynamic errors that are thrown when an XPath expression refers to the context when the context item is undefined (maybe differentiated by the cause of the undefined context item).

ndw:
It's not static analysis; it's a runtime constraint. (You could sometimes guarantee that it won't happen, but you can't detect if it will or not until you're running the pipeline.)

1. If you have a value template or `p:with-option` or `p:variable` expression that doesn't refer to the context item (`1+2`, `$foo`, etc.) then it doesn't matter what the context is. You'll never have a problem.
2. If you refer to the context item then you will get an error if the provided context does not consist of exactly one item. It's a quality of implementation issue to make that message meaningful.
3. My proposal says that if you read from a port (implicitly in the case of value templates or explicitly), then the context item will be undefined if zero items are available on that port or if more than one item is available.

Having given it some more thought, I think point 3 is really just an implementation strategy. We only have to say that it's an error. How I detect that error isn't really relevant.

But the critical point is that you do not get first value semantics.

By the way, if you know you're getting a sequence, you can use collection=true on the long form of the with-option or variable, so there's no loss of functionality.

eriksiegel:
Ok, I think I can explain this to the audience. Let's do it this way.

I think I do need some additional explanation about the meaning of `@collection` (that I completely mis-interpreted probably). But that's better done outside this discussion thread.

ndw:
So can I merge this now? @xml-project did I address your concerns (I can't work out from the GitHub UI what the text of your comment is/was!)

xml-project:
@ndw Everything is fine with me. I originally approved the PR but then reverted it because of @gimsieke intervention. This is why there is no comment!

PR #567, opened 2018-10-18
Fixing step's signature of p:cast-content-type

### LINKED ISSUE #566: Signature of "p:cast-content-type" needs a fix
In "p:cast-content-type" 
1. option "content-type" is not mark as required.
2. option "parameters" must is optional, so  @as have to be "map(*)?"

xml-project:
Attempt to fix #566

### DIFF
--- a/steps/src/main/xml/steps/cast-content-type.xml
+++ b/steps/src/main/xml/steps/cast-content-type.xml
@@ -7,8 +7,8 @@ of its input.</para>
 <p:declare-step type="p:cast-content-type">
    <p:input port="source" content-types="*/*"/>
    <p:output port="result" content-types="*/*"/>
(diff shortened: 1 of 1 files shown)


xml-project:
@gimsieke Thanks for the hint: Will check and ask @ndw in case of emergency.

ndw:
The `steps.xpl` file (http://spec.xproc.org/master/head/steps/steps.xpl) is build automatically from the descriptions in the spec.

xml-project:
@ndw So there is no RNG or RNC I have to change, if I change the step's signature?

ndw:
Correct. If you look in your build/dist/steps directory, you'll find the correct version built from your spec change.

PR #569, opened 2018-10-21
Fix typos in description of document-properties functions

ndw:
I don't believe these will be controversial changes, but they will break some tests.
1. I assume that making the output a `p:` document and not a `c:` document was a simple oversight.
2. I clarified that `p:document-properties-document` behaves the same way as `p:document-properties` if the item passed in is not associated with a document.


xml-project:
Remark: I know that approving a PR once is enough. But as GITHUB was down and tells me, it could not process my request, I retried it from time to time, getting always the same answer-
Seems (just looking at the number approvals), to be my favorite PR so far.

ndw:
Yeah, GitHub has been majorly borked for the past day or so: https://status.github.com/messages

PR #577, opened 2018-10-24
Rework description of common attributes

### LINKED ISSUE #29: Standardize cx:message attribute

### LINKED ISSUE #554: Reconsideration of the @[p:]... construct
I'm trying to get my head around the concept that attributes, like `use-when` and `expand-text` are, dependent on the element they're occurring on, are in the `p:` namespace or not. And some, like `p:timeout` are always in the `p:` namespace... 

In the specs there's also the `depends` attribute which is described (although not explicitly) as always *not* in the `p:` namespace.

I'm sure there was a reason for all this (I forgot but I remember talking about this to Norm), but it is a rule that rather complicates things for, probably?, an edge case. 

Can't we simply decide that *all* standard attributes (and options) XProc introduces are in the `p:` namespace? I think future XProc users will be rather grateful for this simplification so they don't have to look up what type (always `p:`, no-namespace, dependent on element namespace) a standard attribute is.

ndw:
This my attempt to implement what we decided to do about 554. I believe it also closes 29.

Fix #554 
Fix #29

### DIFF
--- a/src/main/schema/core30.rnc
+++ b/src/main/schema/core30.rnc
@@ -124,11 +124,55 @@ expand-text.attr = attribute expand-text { xsd:boolean }
 ]
 p_expand-text.attr = attribute p:expand-text { xsd:boolean }
 
+[
+   sa:ignore = "yes"
+]
+inline-expand-text.attr = attribute inline-expand-text { xsd:boolean }
+
+[
+   sa:ignore = "yes"
+]
+p_inline-expand-text.attr = attribute p:inline-expand-text { xsd:boolean }
+
+[
+   sa:ignore = "yes"
+]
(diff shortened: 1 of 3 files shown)


xml-project:
@ndw We broke it. In my PR from yesterday I moved "common attributes" one level up, so there would be two sections in this document.

With your PR you remove to section completely and now there is a conflict. I do not dare trying to resolve it (because I have never done this be before). Can you please fix it?

xml-project:
@ndw Thanks!

Issue #553, opened 2018-10-07
Type of document-properties


ndw:
It's currently `map(QName,item())`, it's uncontroversial that it is meant to be `map(QName,item()*)`,
right?

xml-project:
As I read our discussions: Yes!

PR #586, opened 2018-10-30
Enhance document-property() function

### LINKED ISSUE #585: Enhance p:document-property() function?
Since we've changed the document properties into a `map(xs:QName, item()*)` it will be a nuisance to get anything out of this map because you always have to use the QName function to create a key and get something out. So I think it's a good idea to enhance the p:document-property() function:

Now: `p:document-property($doc as document-node(), $key as xs:QName) as item()` (the return type is wrong anyway I think)

Proposed: `p:document-property($doc as document-node(), $key as item()) as item()*`

Rules for interpreting $key are roughly the same as for p:force-qname-keys():

- If $key is of type xs:QName it is used unchanged
- If $key is of type xs:string it is transformed into an xs:Qname using the XPath EQName production rules. That is, it can be written as a local-name only, as a prefix plus local-name or as a URI plus local-name (using the Q{} syntax). It is a dynamic error (err:XD0061) if a string $key cannot be converted into a QName.
- If $key is of any other type, the function returns the empty sequence.

eriksiegel:
This resolves #585 and also fixes some typos

### DIFF
--- a/xproc/src/main/xml/specification.xml
+++ b/xproc/src/main/xml/specification.xml
@@ -1819,15 +1819,36 @@ normally during static analysis.</para>
 <glossterm>document properties</glossterm> of a document.</para>
 
 <methodsynopsis>
-<type>item()</type>
+<type>item()*</type>
 <methodname>p:document-property</methodname>
 <methodparam><type>document-node()</type><parameter>doc</parameter></methodparam>
-<methodparam><type>xs:QName</type><parameter>key</parameter></methodparam>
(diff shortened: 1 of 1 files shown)


eriksiegel:
I inadvertently also resolved #553 👍

eriksiegel:
@xml-project Choice for `item()` was inspired by `p:force-qname-keys()` that also works on a key of type `item()` and has roughly the same rules for treating it. You can throw in anything and it simply ignores what it doesn't understand. 

So main reason for choosing `item()`: Consistency with previously made choices.

Issue #580, opened 2018-10-24
Signature of p:error


xml-project:
The current signature of p:error is:
````
<p:declare-step type="p:error">
     <p:input port="source" sequence="true" content-types="*/*"/>
     <p:output port="result" sequence="true" content-types="*/*"/>
     <p:option name="code" required="true" as="xs:QName"/>         
     <p:option name="code-prefix" as="xs:NCName"/>                 
     <p:option name="code-namespace" as="xs:anyURI"/>              
</p:declare-step>
````
So any type of document can appear on p:input. But as the example suggests, the error document is created by wrapping a c:error-element around the content of port "source". What does that mean:

1. For xml media type this description works.
2. For text media type it could mean to add a text node to c:error.
3. For JSON media type: No idea.
4. For other media type: Makes no sense.

Proposal: 
Restrict content types of "source" to XML media types and Text media types.

xml-project:
Fixed with PR #592

PR #592, opened 2018-10-30
Attempt to fix-580

xml-project:
Fixing #580 ,
correcting attributes to "empty-sequence" is default rule and
adapt to changes for construction of QName-attributes.


xml-project:
@ndw Could you please check the build pipeline. If I understand the error raised by Travis right, "xs:anyAtomicType" is not allowed for "as".
Could you give me a hint, what to do instead.

ndw:
Rebase this PR with the fix I just merged into master and you should be fine. I had to patch jing :-(

See #598

xml-project:
Back in a minute (or two)

PR #600, opened 2018-10-31
Attempt to fix-580

xml-project:
Changed signature and adapted text to QName manufacturing


xml-project:
Now it works. Thanks @ndw !

PR #604, opened 2018-10-31
Rework section 3.1, document types

### LINKED ISSUE #588: Add JSON as a specific document type + fix naming?
Section [3.1](http://spec.xproc.org/master/head/xproc/#document-types) of the spec mentions XML, text and non-XML as document types. The following remarks:

1. Shouldn't JSON be mentioned as a specific type? I think (not sure) we decided that JSON "flows" through the pipelines as an XDM type. We discussed this briefly in #555, specifically [this comment](https://github.com/xproc/3.0-specification/issues/555#issuecomment-429271382) mentions what you get when you use `collection()` to retrieve a JSON document. 
1. *non-XML* as document type name IMHO doesn't make much sense any more... Text is also non-XML. Change to *binary* or *other*?

ndw:
Fix #588.

### DIFF
--- a/xproc/src/main/xml/specification.xml
+++ b/xproc/src/main/xml/specification.xml
@@ -364,21 +364,69 @@ defined properties.</para>
 <section xml:id="document-types">
 <title>Document Types</title>
 
-<para>From an XProc perspective, there are two kinds of documents: XML
-documents and non-XML documents. Non-XML documents can be further subdivided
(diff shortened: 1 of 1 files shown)


ndw:
Let me do a little updating.

ndw:
You can see the formatted text here: http://xpspectest.nwalsh.com/iss-588/head/xproc/#document-types

Issue #581, opened 2018-10-25
err:XS0084 is still wrong


ndw:
Surely this:

> It is a static error (err:XS0084) if the p:inline-expand-text attribute appears on any element in the XProc namespace that is not a descendant of a p:inline element or implicit inline. 

should say "inline-expand-text" not "p:inline-expand-text".

ndw:
Fixed by #607, I think.

PR #607, opened 2018-10-31
Add static error for p: attributes on p: elements

### LINKED ISSUE #582: We need a static error for putting p:expand-text and friends on elements in the p: namespace
I don't think we have one yet...

ndw:
Fix #582 

1. This PR adds a static error for p: attributes on p: elements
2. It also fixes what I assume was a carelessly structured description of err:XS0084.


ndw:
Fix #581 
This PR also fixes this issue, I believe.

PR #608, opened 2018-10-31
Clarify serialization in p:document-properties-map

### LINKED ISSUE #573: Format returned by p:document-properties-document
Now that document properties are `map(xs:QName,item()*)` we need to update the format of the XML document returned. We currently do not cover the case where more than one item is associated with a key.

ndw:
Fix #573 
Ths PR attempts to clarify how `p:document-properties-document` serializes the properties. This is a convenience function to make processing properties with pipeline steps easier. I think that complicated property values are going to be uncommon and perfect fidelity is not a goal of this function. But you may disagree :-)

### DIFF
--- a/xproc/src/main/xml/specification.xml
+++ b/xproc/src/main/xml/specification.xml
@@ -1818,13 +1818,19 @@ repeated for each item in the sequence.
 </para>
 </listitem>
 <listitem>
-<para>If a value is a map, the element content is the JSON serialization
-of that map.
+<para>If a map appears as a value, the property element will have an
+attribute named <tag class="attribute">map</tag> with the value “<code>true</code>”.
+<impl>It is <glossterm>implementation-defined</glossterm> how or if a
+processor attempts to serialize maps that appear in document properties.</impl>
+If the implementation does not serialize maps, the property element will be empty.
 </para>
 </listitem>
 <listitem>
-<para>If a value is an array, the element content is the JSON serialization
-of that array.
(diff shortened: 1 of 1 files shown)


ndw:
Sure, we can talk about it this afternoon.

gimsieke:
For `xsi:type`, we can link to https://www.w3.org/TR/xmlschema11-1/#xsi_type.
I see that its value must be a QName, therefore it will be hard to use something like `as="map(*)"` for `xsi:type`. Instead of `xsi:type`, we could use an `as` attribute that holds a [sequence type](https://www.w3.org/TR/xpath-31/#dt-sequence-type), or we could invent `xsi:type="map:any"` with [`xmlns:map="http://www.w3.org/2005/xpath-functions/map"`](https://www.w3.org/TR/xpath-31/#dt-qname) (similarly, `array:any`).
Why did we use `xsi:type` instead of `as` (sequence type) anyway?

ndw:
Implementation-defined is fine by me.

I proposed xsi:type because it's a document that we're producing and that's the standard way to represent types in XSD. It does present a problem for arrays and maps though. Writing XPath expressions to decompose a complicated sequence type isn't going to be fun. And I'm not sure about the consequences of inventing our own type QName. If someone runs that through a process that attempts to do any kind of validation, it's going to fall over because there won't be an XSD declaration for the type.

We could just stipulate "`map='true'`" or "`array='true'`" for maps and arrays.

ndw:
Right. I made it implementation defined.

PR #617, opened 2018-11-02
Merge steps-intro prose into the other specs; remove it

### LINKED ISSUE #612: Move "QNames as option values" into the core spec; kill the step introduction document

ndw:
Fix #612


ndw:
I believe I've put back all the errors. I've rewritten the section a fair but, no offense intended.

PR #621, opened 2018-11-02
Attempt to fix 597

xml-project:
Removed requirement that the p:catch clauses have to be considered in document order. The only order requirement now is, that the p:catch without a code-attribute has to be the last.
I think that is reasonable and will enforce good coding practice, but I am open to the argument, that should just enforce that there is only one p:catch without code-attribute, wherever it may appear in the sequence.


ndw:
I definitely think we should require the catch without a code to be last.

PR #632, opened 2018-11-09
Fix description of non-step wrapper for p:try

xml-project:


xml-project:
@ndw Not sure you are right here. To my reading this passage is about defining what kind of steps 'p:when', 'p:otherwise', 'p:catch' or 'p:finally' are. It is already stated that 'p:try' is a multi-container step with an initial subpipeline. I am not sure we really need an additional concept "implicit non-step wrapper".
What did I miss?

xml-project:
Closed in favour of #636

PR #635, opened 2018-11-10
Attempt to fix 634

### LINKED ISSUE #634: XS0044 or XS0100
Now that we have XS0100:

> It is a static error (err:XS0100) if the pipeline document does not conform to the grammar for pipeline documents. This is a general error code indicating that the pipeline is syntactically incorrect in some way not identified more precisely in this specification.

I propose to restrict X0044 to the mentioned special case:

> It is a static error (err:XS0044) if any element in the XProc namespace or any step has element children other than those specified for it by this specification. In particular, the presence of atomic steps for which there is no visible declaration may raise this error.

xml-project:
Attempt to fix #634 by making it an error for the special case of an undeclared atomic step.

### DIFF
--- a/xproc/src/main/xml/specification.xml
+++ b/xproc/src/main/xml/specification.xml
@@ -3476,10 +3476,8 @@ attribute">extension attributes</glossterm>.</error>
               value does not satisfy the type required for that attribute.</error></para>
         </listitem>
         <listitem>
-          <para><error code="S0044">It is a <glossterm>static error</glossterm> if any element in
-              the XProc namespace or any step has element children other than those specified for it
-              by this specification. In particular, the presence of atomic steps for which there is
-              no visible declaration may raise this error. </error></para>
+          <para><error code="S0044">It is a <glossterm>static error</glossterm> if any step contains 
+            an atomic step for which there is no visible declaration.</error></para>
         </listitem>
         <listitem>
           <para><error code="S0037">It is a <glossterm>static error</glossterm> if any user extension 


xml-project:
@gimsieke IMHO xs:0044 should cover
````
<p:inserter ..... /> <!-- misspelled step name -->
<ab:not-definied />
````
So moving it just to  http://spec.xproc.org/master/head/xproc/#p.extension does not cover all cases. Or do you suggest just to repeat it in the referenced section?

gimsieke:
That’s true. However: One may argue that S0010 (“It is a static error if a pipeline contains a step whose specified inputs, outputs, and options do not match the signature for steps of that type.”) does not belong to the extension steps section alone, but also to the [atomic steps](http://spec.xproc.org/master/head/xproc/#p.atomic) section.
I wouldn’t suggest repeating it (apart from in appendix E). We can leave it where it is, but then we should consider moving S0010 to the common errors.

Another solution would be to regroup the current sections 16.8 and 16.9:
16.8 Atomic Steps
16.8.1 Standard Atomic Steps
16.8.2 Atomic Extension Steps
16.8.3 Atomic Step Errors

xml-project:
Sorry, I am lost. In my version XS0100 is

> It is a static error (err:XS0100) if the pipeline document does not conform to the grammar for pipeline documents. This is a general error code indicating that the pipeline is syntactically incorrect in some way not identified more precisely in this specification.

and is in the common error section. 
Any help appreciated!

gimsieke:
This regrouping will have the benefit that it eliminates ambiguity when we speak of atomic steps, sometimes referring to both standard and extension steps (as we do in error S0044), sometimes only to steps in the XProc namespace.
For example, the first para of 17.5, [p:declare-step](http://spec.xproc.org/master/head/xproc/#p.declare-step), might need some clarification:
> A p:declare-step provides the type and signature of a pipeline or an atomic step. Pipelines contain a subpipeline which defines what the declared step does. Atomic steps have an implementation defined elsewhere in some other way.

The last sentence probably refers to standard atomic steps only?
Also 17.5.2: 
> The distinction between an atomic step declaration and a pipeline declaration hinges on the presence or absense of a subpipeline. An atomic step declaration does not have a subpipeline. 

If “atomic step” as a collective term refers to both standard and extension atomic steps, we need to give up Sect. 17’s distinction in atomic steps and pipelines by also calling them “standard and extension atomic steps”. 

When talking about extension steps, one needs to also make a distinction between user-supplied steps written in XProc and processor extension steps. Like standard atomic steps, processor extension steps don’t have a subpipeline in their `p:declare-steps`. 

This discussion probably needs to take place elsewhere, not in the comments to a pull request that is only tangentially related. 
Is it just me who thinks the current terminology is inconsistent?

gimsieke:
> Sorry, I am lost. In my version XS0100 is

I wasn’t referring to S0100 but to S0010.

xml-project:
> I wasn’t referring to S0100 but to S0010.

Sorry: Misreading the code. What do you suggest now? If I merge this pull request your general remarks get lost. If I close it and we decide to discuss the changes you requested first, they are also lost.

gimsieke:
Merge it and I’ll create a new issue that links to the comments

PR #638, opened 2018-11-13
Support new #types and values attribute

### LINKED ISSUE #534: Non-required options and @as

### LINKED ISSUE #613: Consider a mechanism for defining enumerated values for options

ndw:
(This is #637 refactored to use #type instead of p:type)

Fix #534
Fix #613

Support new types (#XSLTSelectionPattern, #XPathExpression, #XPathSequenceType, #RegularExpression, and #QName). Support new values attribute for enumerating values.

Updates step signatures to make optional options have a type that would allow empty sequence.

Tweaked schemas and tools as necessary.

### DIFF
--- a/src/main/resources/css/xproc.css
+++ b/src/main/resources/css/xproc.css
@@ -1,3 +1,11 @@
+p code {
+    font-size: 120%;
+}
+
+p.element-syntax code {
+    font-size: 100%;
+}
+
 dl.toc             { margin-top: 0;
                      margin-bottom: 0
 }
--- a/src/main/schema/core30.rnc
+++ b/src/main/schema/core30.rnc
@@ -8,19 +8,24 @@ namespace xs = "http://www.w3.org/2001/XMLSchema"
 # This schema neither supports nor requires the use of RELAX NG DTD Compatibility
 
 [
-   sa:model = "XPathExpression"
+   sa:model = "#XPathExpression"
 ]
 XPathExpression = text
 
 [
-   sa:model = "XSLTSelectionPattern"
+   sa:model = "#XSLTSelectionPattern"
 ]
 XSLTSelectionPattern = text
 
 [
-   sa:model = "XPathSequenceType"
+   sa:model = "#XProcSequenceType"
 ]
-XPathSequenceType = text
+XProcSequenceType = text
+
+[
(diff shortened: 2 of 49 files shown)


ndw:
In some ways, I think this is better. In some ways, it's more of a hack.

On the whole, I still have a marginal preference for simply saying that attributes of type QName have to be specified as QNames and removing all this magic. But I'd get rid of prefix-value and namespace-value at the same time and that didn't meet with support, so nevermind me.

xml-project:
Yes, I think this is much better and should be much easier to implement on top of Saxon. 
I hesitate to approve this PR in the light of Erik's point about QName conversion we will talk about on Thursday. 
If we decide to do no conversion on call (but only when the value is used), I would like to discuss whether it makes sense to informally treat #QName as a union type of xs:string or xs:QName.

xml-project:
> On the whole, I still have a marginal preference for simply saying that attributes of type QName have to be specified as QNames and removing all this magic. But I'd get rid of prefix-value and namespace-value at the same time and that didn't meet with support, so nevermind me.

The main reason for QName-magic was the user request not having to write `xs:QName("pre:name")` but simply write `pre:name` and let the processor do the rest. As long as they want to write it sometimes as a string, sometimes as a xs:QName, we have to do the magic. 
Covering the legacy features like attribute-prefix and attribute-namespace does not seems like a big deal to me, compared to the (type) magic with #QName.

ndw:
Right. I'm saying that when I look at the scope of this hack, I am a little bit tempted to say, "look, if you want to use XProc 3.0, you're going to have to type `attribute-name="{xs:QName('pre:name')}"` and not `attribute-name="pre:name"`. Life is hard.

xml-project:
:-))

ndw:
Nope. Not this one.

Issue #645, opened 2018-11-23
Names on p:when violate normal scoping rules


ndw:
I don't (think I) object to putting names on `p:when` elements, but note that this does introduce a need to say something about "non step wrappers" or something equivalent.

Normal scoping rules don't apply since, for example, `p:otherwise` must not be able to read from the preceding sibling `p:when` "steps" but now that they have `name` attributes, it would be possible to write `p:pipe` elements that did so.

xml-project:
Sorry to cause problems, but where is the difference between non-step wrappers p:when/p:otherwise on one hand and p:catch/p:finally on the other hand. To my understanding with respect to scoping, both are the same. So why can p:catch/p:finally have names (and cause no trouble) and p:when/p:otherwise not.

ndw:
Sorry. I wasn’t trying to suggest it was different. I was just reminded that it’s an issue we have to tackle.

Not having names on p:when may have been an effort to avoid a problem, but I can see how it would make some things more difficult.

xml-project:
Another though: In XProc 1.0 p:when/p:otherwise did not have a name, but p:catch had. The reason for this seems to be the fact that input port "error" on p:catch is not primary. To access this port, one has to write `p:pipe step="catch-name" port="error"`. Following this logic, p:finally in XProc 3.0 has to have a name too.
What if we make port `error` a *primary input port* for p:catch/p:finally? There would be an auto-connection to the primary input port of the respective subpipeline. No need to pipe the port anymore and therefor no need for p:catch/p:finally to have a name.
Con: This change would prevent XProc 1.0 pipelines using the pipe from running in XProc 3.0.
Did I manage to expose my point?

ndw:
Fixed in #650.

PR #650, opened 2018-11-30
Updates to p:choose

### LINKED ISSUE #647: Evaluation in p:choose
> The p:choose considers each subpipeline in turn and selects the first (and only the first) subpipeline for which the guard expression evaluates to true in its context. If there are no subpipelines for which the expression evaluates to true, the default subpipeline, if it was specified, is selected.
After a subpipeline is selected, it is evaluated as if only it had been present.

I think some addition is necessary here for the attribute "test" on p:when. Suppose there is an XPath-error in a when-expression which is located after the first when-expression that evaluates to true. Is a processor required to raise an error (which can be caught) or not?

ndw:
1. Make it explicit that after a `p:when` is selected, no subsequent test expressions are evaluated. Fix #647 
2. Make it explicit that `p:otherwise` is optional and describe what happens if it isn't present.
3. Simplify the content model of `p:choose` in the schema.

Right. So it's possible that item 3 will be controversial. The previous content model guaranteed that a `p:choose` would contain at least one `p:when` or `p:otherwise`. However, we have XS0074 that doesn't appeal to the schema so it isn't necessary. I think the new content model is sufficiently simpler that it's worth losing the grammar constraint. But I won't argue if anyone disagrees.

### DIFF
--- a/src/main/schema/core30.rnc
+++ b/src/main/schema/core30.rnc
@@ -580,12 +580,16 @@ Choose =
       common.attributes,
       global.attributes,
       step.attributes,
-      (Documentation|PipeInfo)*,
-      WithInput?,
-      (Documentation|PipeInfo|When)*,
-      (Documentation|PipeInfo)*,
-      Otherwise?,
-      (Documentation|PipeInfo)*
+
+      ((Documentation|PipeInfo)*,
+       WithInput?,
+
+       ((((Documentation|PipeInfo)*, When)+,
+         ((Documentation|PipeInfo)*, Otherwise)?)
+       | (((Documentation|PipeInfo)*, When)*,
+         ((Documentation|PipeInfo)*, Otherwise))),
+
+       (Documentation|PipeInfo)*)
    }
 
 [
--- a/src/main/schema/core30.rnc
+++ b/src/main/schema/core30.rnc
@@ -580,15 +580,12 @@ Choose =
       common.attributes,
       global.attributes,
       step.attributes,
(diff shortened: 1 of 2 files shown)


ndw:
I wanted to fix #645 as well, but it was going to have conflicts with this PR, so I just added the fixes to this PR. Not ideal, but seemed better than waiting for approval of this one before proceeding.

ndw:
I agree that `p:finally` is always executed.

I'll put the content model back. *shrug*.

Like I said, error XS0074 says that at least one of when or otherwise is required; we don't need to enforce that in the grammar. But the grammar isn't so complicated that I think it matters.

ndw:
Actually, we already say:
> Irrespective of whether the initial subpipeline succeeds or fails, if any recovery pipelines are selected, and whether they succeed or fail, the p:finally block is always run after all other processing of the p:try has finished.
But since I'm mucking about, I'll try to clarify that.

ndw:
I did not mean to close this one. Fumble-fingered the button, I guess.

xml-project:
> Irrespective of whether the initial subpipeline succeeds or fails, if any recovery pipelines are selected, and whether they succeed or fail, the p:finally block is always run after all other processing of the p:try has finished.

Yes, you are right. Reading this passage in our context it is clear, that p:finally is always executed. Skipped this. Sorry!

ndw:
No worries. I missed it too on a quick reading, so I've pulled it out into a big long list of cases with clear descriptions (I hope) of what happens and what errors must be reported.

PR #651, opened 2018-11-30
Reorganize steps spec

ndw:
I'm not sure why I got a bee in my bonnet a couple of weeks ago about the os and file steps. Maybe I was looking for something mechanical to do. Anyway, I've refactored those steps out into separate specs, because I think that's almost certainly what we intended.

Along the way, I took the liberty of tinkering with `p:file-head` and `p:file-tail` to make them actual filters (operating on an input `text/plain` document and not a URI.


ndw:
Maybe I imagined it.

But they are collections of closely related steps that might or might not be implementable. Suppose you're doing an XProc implementation for some embedded system which doesn't have a filesystem. What then? Suppose you're implementing it on some mainframe that doesn't have environment variables. What then?

xml-project:
Strange places for an XProc processor, but this is obviously an argument to make these steps optional. .... On the hand: What about p:http-request, p:load, p:store and p:document? Shouldn't they be optional as well following your argument.
Concerning missing environments: If the mainframe running an XProc processor does not have environment variable, I would argue, that a `c:result`-document with no children should be returned.

ndw:
¯\_(ツ)_/¯

We could put all of the steps in separate specs. We could put all of the steps in the same spec. I don't really understand the rational that we're applying. Putting the validation steps and the XSL formatter step and `p:exec` and `p:run` in separate specs but putting all of the file and os steps in the "main" spec bothers my aesthetic sensibilities, but I surrender.

gimsieke:
On systems without filesystems or environment variables these steps could simply return a document with an empty `c:result` element (or without a `cwd` element/attribute in the case of `p:os-info` and lacking filesystem). This is independent of whether the steps are optional or not. But this would allow the steps to remain mandatory.

xml-project:
You do not have to surrender, because this is an important point, we need to settle. To my understanding we have one document for the required steps (standard library) and more than one (I do not think there is a rule) for all optional steps. I do not care about the question whether all file steps are in one document or in say ten. I just argued they should be required steps.

PR #669, opened 2018-12-11
remove caveat as the patch is now accepted

innovimax:
relaxng/jing-trang#229


ndw:
Works for me. Thanks.

PR #676, opened 2018-12-26
A round of editorial changes

ndw:
Over Christmas, I tried to do an end-to-end read through and make some editorial improvements. I haven't yet got all the way through, but here's a first bunch.

I don't think there's anything likely to be controversial in here, or viewed as non-editorial. Certainly that wasn't my intent.

The most striking change is probably that I redid the graphics because they were awful. (I can say that because I'm responsible for the ones in the 1.0 spec.) I also added a diagram at the beginning of section 6 which (I hope) clarifies what constitutes a connection.

Comments, etc.most welcome.

I've just discovered that diffs aren't working again. Will fix that later today.


eriksiegel:
I suppose there is a way to see the end result as a web page?

ndw:
Of course: http://xpspectest.nwalsh.com/editorial/head/xproc/

PR #683, opened 2019-01-03
Attempt to fix #451

### LINKED ISSUE #451: Introduce a file system path → URI normalization function?
In the light of the discussions in #325 and #449, I’d like to propose a function `p:urify()` as described in https://github.com/xproc/3.0-specification/issues/325#issuecomment-374000201.

It should basically convert file system paths to `file:` URIs.

This function is useful because users often like to submit file system paths instead of URIs as options, while most steps with an `href` or similar option only accept URIs.
The issue has become more pressing because with #449 we introduced space-separated `href` values. When heuristically accepting file system paths instead of proper URIs, a processor must not interpret spaces in the argument as part of the (file system) name. Instead, each space must be regarded as separating (possibly relative) URIs. 
We should exonerate users from the burden of percent-escaping, `file:`-prepending, … file system paths.

`p:urify()` should by and large be idempotent if the argument is already recognized as an absolute URI. It could, however, perform additional normalizations on URIs such as eliminating consecutive duplicate slashes or, if the argument is determined to represent a directory, append a trailing slash.

When passing user-supplied paths to a step that accepts space-separated URIs, a pipeline author could just apply `p:urify` to the paths:

```xml
  <p:document href="{$paths ! p:urify()}"/>
```

(assuming that a sequence of URIs will be flattened to a space-separated string that is then tokenized and each token is cast as a URI). 

Relative paths should be resolved to the current working directory (as currently returned by [`pos:info`](http://exproc.org/proposed/steps/os.html), not to the pipeline’s base URI. 

Path separators should be interpreted as per the system property reported by `pos:info`, with the exception that if the separator is reported as `\`, `/` should also be interpreted as a separator.

We have implemented a similar function as a step, [`tr:file-uri`](https://github.com/transpect/xproc-util/tree/master/file-uri/xpl/file-uri.xpl). It does a bit more than the suggested function, namely optional catalog resolution and optional retrieval of HTTP resources (stored in a temporary directory). It returns an XML structure with the local URI, the current working directory, the last path component, etc. 

I see some value in having a function though because it can be used in AVTs and XPath expressions in general, without the need to insert another step into the subpipeline. 

It would still be nice to be able to use a catalog resolver even for URIs that don’t refer to XML resources. Maybe `p:urify()` can accept an optional parameter map that provides implementation-defined functionality, such as `{ 'use-xml-catalogs': true(), 'fetch-http': true() }`.

gimsieke:
Specifying a `p:urify()` function whose details remain _implementation-dependent_.

close #451

### DIFF
--- a/xproc/src/main/xml/specification.xml
+++ b/xproc/src/main/xml/specification.xml
@@ -2118,7 +2118,7 @@ normally during static analysis.</para>
     For Windows paths, the forward slash and the backslash <rfc2119>should</rfc2119> be considered equivalent. Operating
     systems with other filesystem path addressing schemes (for example, VMS or Mac OS) need not be supported.</para>
   <para>Each argument may be an operating system path, including paths with drive letters and UNC paths on Windows, or a URI.
-    On non-Windows systems, file URIs that contain an authority component (<literal>file://hostname/path</literal> may be reported 
+    On non-Windows systems, file URIs that contain an authority component (<literal>file://hostname/path</literal>) may be reported 
     as an error.</para>
   <para>The function need not make sure that an argument identified as a URI is already a <emphasis>correct</emphasis> URI
     according to <biblioref linkend="rfc2396"/>. It may attempt to correct errors in purported file URIs, such as fixing the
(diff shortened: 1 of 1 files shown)


gimsieke:
A sample implementation is [here](https://github.com/xproc/Workshop-2018-09/blob/master/p_urify/test.java)

PR #688, opened 2019-01-03
Fix markup issues in p:urify()

ndw:
This PR just makes a few small markup and editorial improvements to the p:urify() function.

1. In order to gather implementation defined/dependent bits in an appendix, you have to wrap the statement in an `<impl>` tag.
2. In order to make the bits in the appendix understandable, you have to write them as sentences that will stand alone, out of context.
3. I fixed the section title so it's more in line with the others.


ndw:
Sorry, I also made one ostensibly non-editorial change. I made the algorithm used implementation-defined rather than implementation-dependent.

PR #695, opened 2019-01-06
Remove multi-step container, non-step wrapper, and related terms

### LINKED ISSUE #39: Legacy from 1.0: # 3.1 Editorial improvements: non-step wrappers
# 3.1 Editorial improvements: non-step wrappers 

>This is an aggregation of the discussion on "3.1 Editorial improvements: non-step wrappers" ([Issue 45 of the xproc/1.0-specification](https://github.com/xproc/1.0-specification/issues/45))

**Opened by: ndw on 2014-08-22, 18:38h**

## ndw said on 2014-08-22, 18:38h:

Implementation experience has demonstrated that there are areas of the specification that didn't get the balance right between precision for implementers and clarity for users, for example “non-step wrappers”. The XProc 2.0 specification should attempt to resolve these problems without introducing inordinate complexity. 

The 1.0 specification also defines the p:pipeline element as a syntactic shortcut for a particular form of p:declare-step. While convenient in some circumstances, it has proven to be a source of some confusion especially among new users. XProc 2.0 may remove the p:pipeline element. 

	On 2014-08-22, 18:38h: ndw added requirement should labels.

	On 2014-08-25, 22:28h: ndw added this to the XProc 2.0 LC milestone.

	On 2015-10-07, 15:04h: ndw modified the milestone: XProc 2.0 Next PWD, XProc 2.0 LC.

### LINKED ISSUE #636: Rethink non-step wrappers
Currently the [specs](http://spec.xproc.org/master/head/xproc/#step-concept) say:

> Each subpipeline is identified by a non-step wrapper element: p:when and p:otherwise in the case of p:choose, p:group and p:catch in the case of p:try.

Here we have at least to remove the p:group from the p:try-list and add a p:finally. But as @ndw [remarked](https://github.com/xproc/3.0-specification/pull/632):

> I'm not sure what to do about the whole "non-step wrapper" terminology. But if we're going to keep it, I think you have to say something about the fact that the body of the p:try element preceding the first p:catch is somehow in an implicit non-step wrapper.

ndw:
Fix #39 and fix #636 

It was a lot harder to figure out how to do this than I expected, but I believe this PR removes the terms "multi-step container", "non-step wrapper", and related complexities without doing harm to the spec. I also think the definition of container has been improved.

### DIFF
--- a/xproc/src/main/xml/specification.xml
+++ b/xproc/src/main/xml/specification.xml
@@ -70,11 +70,13 @@ pipeline itself, or from the outputs of other steps in the pipeline.
 The outputs from a step are consumed by other steps, are outputs of
 the pipeline as a whole, or are discarded.</para>
 
-<para>There are three kinds of steps: atomic steps, compound steps,
-and multi-container steps. Atomic steps carry out single operations
-and have no substructure as far as the pipeline is concerned. Compound
-steps and multi-container steps control the execution of other steps,
-which they include in the form of one or more subpipelines.</para>
+<para>There are two kinds of steps:
+<glossterm baseform="atomic step">atomic steps</glossterm> and
+<glossterm baseform="compound step">compound steps</glossterm>.
+Atomic steps carry out a single operation and have no substructure as
+far as the pipeline is concerned. Compound steps control the execution
+of other steps, which they include in the form of one or more
+subpipelines.</para>
 
 <para><biblioref linkend="steps30"/>
 defines a standard library of steps. Pipeline implementations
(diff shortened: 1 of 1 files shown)


ndw:
Here's the formatted version, if it helps: http://xpspectest.nwalsh.com/iss-39-636/head/xproc/

I've reached out to the Delta XML guys for help in getting diffs working again.

xml-project:
>Here's the formatted version,

Yes, this makes it a lot easier

PR #699, opened 2019-01-08
Address two extant FIXME: comments in the spec

ndw:
The first is practically editorial, the second is `p:import-functions`!

How we managed to close the issue about that before we wrote any prose, I couldn't say. This is a bit minimal, but at least it beats "FIXME".

### DIFF
--- a/src/main/schema/core30.rnc
+++ b/src/main/schema/core30.rnc
@@ -225,7 +225,7 @@ ImportFunctions =
       global.attributes,
       href.attr,
       content-type.attr?,
-      attribute namespace { xsd:anyURI }?,
+      attribute namespace { xsd:anyURI+ }?,
       (Documentation|PipeInfo)*
    }
 
--- a/src/main/schema/core30.rnc
+++ b/src/main/schema/core30.rnc
@@ -224,6 +224,8 @@ ImportFunctions =
       common.attributes,
       global.attributes,
       href.attr,
+      content-type.attr?,
+      attribute namespace { xsd:anyURI }?,
       (Documentation|PipeInfo)*
    }
 
--- a/tools/xsl/rngsyntax.xsl
+++ b/tools/xsl/rngsyntax.xsl
@@ -245,6 +245,14 @@
 	<xsl:when test="rng:data">
 	  <xsl:apply-templates/>
 	</xsl:when>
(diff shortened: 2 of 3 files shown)


ndw:
We can definitely talk about this on Thursday, but a few thoughts now:

> a) For the sake of clarity, I think we need to say, that recursive imports (in the XSLT or XQuery imported) need to be followed.

I think we have to say (somehow) that the imported library is parsed and loaded according to the specifications that govern it. You can't meaningfully load an XSLT or XQuery without parsing it according to its semantics. I think we want to say that without calling out those languages specifically. 

> b) What do we say about competing imports where two imports with the same namespace import a function with the same name? I think we should say it is an error. (Alternative: Take first or last)

Error.

> c) I think we should have a static error saying: I know (in principle) how to deal with these import - as opposed to XS0103), but this import is defect (a syntax error or a circular import).

Yep.

> d) The content type does not specify the version, so a XSLT 3.0 is to be imported, but I only know, how to handle XSLT 2.0. Is this XS0103?

I think so. This is a static error, I don't think pipeline authors are going to have too much trouble understanding "I couldn't load it".

> e) I would like to have a boolean option on p:import-function that say "Raise a static error, if this import fails." This prevents pipeline authors from handling a lot of dynamic errors resulting from unknown function-names in XPath expressions.

Unknown function names in XPath expressions are static errors, surely?

> f) I would like to see the namespace to be xs:anyURI*, so I can select more than one namespace to be imported and do not have to repeat myself for every namespace.

Ok.

> g) I think it does make sense to add the ability to import a certain content-type to p:system-properties, so one can make an import conditional with `use-when".

That is a fabulous idea.

xml-project:
> Unknown function names in XPath expressions are static errors, surely?

Why? I would think they are dynamic *XProc* errors.(Of cause it is a static XPath error) It was XD0023 in XProc 1.0 but we lost that error code somehow. What did I miss?

ndw:
If the content-type attribute is missing, the processor guesses. If the type is provided and is incorrect, that's an error if it's wrong. The processsor is required to use the specified.

ndw:
Fix #704

ndw:
Yes. Exactly right on the missing "not".

PR #709, opened 2019-01-16
A collection of mostly editorial changes

ndw:
I'm about half-way through on my editorial pass. I believe that all of these changes are editorial in the sense that they don't change the technical content of the spec. But a few turned out to be bigger than just typos.

I've tried to avoid reflowing paragraphs and other things to keep the diffs clean(ish). There's also a formatted version here: http://xpspectest.nwalsh.com/editorial/head/xproc/

Diffs are still broken, alas :-(


ndw:
Suggestions welcome. I'll do a second read through after the first.

PR #715, opened 2019-01-17
Attempt to fix 711

### LINKED ISSUE #711: Default settings for serialisation parameters
Currently we [say](http://spec.xproc.org/master/head/xproc/#serialization):

> The default value of any serialization parameters not specified on a particular output is implementation-defined.

Playing a little around with this I would propose, that the processor is required to set default of 'method' according to the document type, so:

- XML document -> method = xml
- JSON document -> method = json
- Text document -> method = text
- Other documents -> Implementation defined.

Does this make sense to anyone?

xml-project:
close #711


xml-project:
What do we do about this PR? Close and come back to the issue later? Any comment welcomed.

PR #729, opened 2019-01-26
Fix another markup problem

ndw:
Sorry I missed this in the earlier PR. There was a markup error (a glossterm'd static-error that wasn't in an error element) and that made me look more closely at the paragraph.

I've reworded it slightly to improve the English.

I have a question though, why is the case of type errors being called out specifically here. We have a rule that says dynamic errors can be raised statically if the processor can prove they will always occur, does that rule not apply to type errors? If not, why not?

### DIFF
--- a/xproc/src/main/xml/specification.xml
+++ b/xproc/src/main/xml/specification.xml
@@ -1013,12 +1013,13 @@ performing a static analysis of all XPath expressions, and detecting static erro
 evaluation</firstterm> consists of tasks which, in general,
 cannot be performed out until a source document is available.</termdef></para>
 
-<para><error code="S0107">It is a <glossterm>static error</glossterm> if any static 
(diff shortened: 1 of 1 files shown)


xml-project:
1. Sorry for the markup-error.
2. Thanks for improving my English. Punctuation is really the worst part of it.
3. I think we need to call out type error in XPath expression as I did, because a conformant XPath processor may either report typing errors statically, or report them dynamically. In the specs this is called "Static typing feature" and its optional.

If we do not call out type error here the behaviour of a pipeline changes if you switch from Saxon EE (static typing feature implemented -> typing error raised statically) to Saxon HE (error raised dynamically).

Did I miss something?

ndw:
No, thanks, I think that's worth highlighting.

PR #731, opened 2019-01-27
Attempt to fix-712 and 713

### LINKED ISSUE #712: What happens when omit-xml-declaration is omitted?
Currently we have (as it was in 1.0, too):

> The omit-xml-declaration must be supported. If the value is not specified or has the value no, an XML declaration must be produced.

In the 1.0 spec, when there was the `omit-xml-declaration` attribute on a standard step, it would always default to `'true'`. Therefore a processor didn’t have a choice but to omit the XML declaration if the option was missing. 

I think now the situation is different if the sentence above holds: There is no default any more for the `omit-xml-declaration` map entry, therefore a 3.0 processor must now produce an XML declaration if `omit-xml-declaration` is not specified.

Do you agree with this interpretation? If so, do we want to have this default behavior (XML declaration will be produced for the `xml` and `xhtml` methods if nothing is specified for `omit-xml-declaration`) that deviates from 1.0?

Another random observation: In the steps spec, we still occasionally refer to XSLT 2.0 and XQuery 1.0 serialization (for ex. in [`p:escape-markup`](http://spec.xproc.org/master/head/steps/#c.escape-markup)). It should probably link to [XSLT and XQuery Serialization 3.1](https://www.w3.org/TR/xslt-xquery-serialization-3/) or to [§ 16.3.1](http://spec.xproc.org/master/head/xproc/#serialization) where we link to the current 3.1 serialization spec. We can fix this while handling the other question raised above.

### LINKED ISSUE #713: Serialization: Default version to '1.0' and encoding to 'UTF-8'
…for the `xml` and `xhtml` methods?

xml-project:
Does this work?

close #712 
close #713

### DIFF
--- a/xproc/src/main/xml/specification.xml
+++ b/xproc/src/main/xml/specification.xml
@@ -4747,11 +4747,11 @@ parameters.
 <para>If serialization method <literal>xml</literal> is choosen either explicitly or
 implicitly, the following default values <rfc2119>must</rfc2119> be used:</para>
 <itemizedlist>
-  <listitem><para>Parameter <literal>method</literal> is set to <literal>1.0</literal>.</para></listitem>
+  <listitem><para>Parameter <literal>version</literal> is set to <literal>1.0</literal>.</para></listitem>
   <listitem><para>Parameter <literal>encoding</literal> is set to <literal>UTF-8</literal>.</para></listitem>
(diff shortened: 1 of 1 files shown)


xml-project:
@gimsieke On Thursdays call we only agreed on default values for version, encoding, and omit-xml-declaration. The other default values are up to the implementation.

PR #797, opened 2019-04-08
Attempt to fix-796

### LINKED ISSUE #796: Different error conditions for XPath expressions
Currently we have:

> XD0008: It is a dynamic error if a sequence of items appears where an item to be used as the context item is expected.

and

>XD0005: It is a dynamic error if more than one document appears on the connection for this input port.

To my reading this means, that
````
<p:variable name="a" select="54">
  <doc /> <doc />
</p:variable>
````
is **not** an error (XD0008) because no context item is expected, but
````
<p:if test="true()">
  <p:with-input><doc /> <doc /></p:with-input>
</p:if>
````
will raise XD0005 even though the context item is not referred.

I fail to see the reason for this difference and think we should remove it.

Did I miss anything?

xml-project:
Attempt to fix #796
Removed XD0008 and changed text of XD0005 a little so it fits both cases.
Also removed remaining "context node" in favour of "context item".

### DIFF
--- a/xproc/src/main/xml/specification.xml
+++ b/xproc/src/main/xml/specification.xml
@@ -3626,16 +3626,16 @@ of the content types that might be produced by any of the <tag>p:when</tag>
 or the <tag>p:otherwise</tag>.
 </para>
 
-<para>The <tag>p:choose</tag> can specify the context node against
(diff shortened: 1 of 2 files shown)


xml-project:
See my comment in issue #796

PR #799, opened 2019-04-23
Added forbidden content types and content type shortcuts

### LINKED ISSUE #790: text/xml no longer XML?
Didn't we decide somewhere that media type `text/xml` is no longer considered XML but text? Can't find it and the spec still says `text/xml` is XML. 

Does anybody have a better memory than me?

ndw:
Fix #790

### DIFF
--- a/xproc/src/main/xml/specification.xml
+++ b/xproc/src/main/xml/specification.xml
@@ -603,6 +603,97 @@ is created and content-type <literal>application/json</literal> is used.</para>
 </itemizedlist>
 </section>
 
+<section xml:id="specified-content-types">
+<title>Specifying content types</title>
+
+<para>In some contexts (step inputs, step outputs, and inline
+elements, for example), XProc allows the pipeline author to specify a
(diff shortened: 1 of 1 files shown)


ndw:
If (when?) this PR is accepted, we should go through the step library and replace the long forms with the short forms.

ndw:
Consider a step like `p:add-attribute` that should apply to XML or HTML documents. You can't say "xml" because that excludes HTML. You can't say "html" because that excludes XML. The temptation will be to say "xml html". That expands to:

> application/xml text/xml application/*+xml -application/xhtml+xml text/html application/xhtml+xml

Without the left-to-right rule, that won't allow `application/xhtml+xml` documents. That was the simplest thing I could think of to make this work.

Alternatively, we could not exclude HTML from XML. Then you wouldn't need the left-to-right rule. I could be persuaded to do that, I guess, but it seemed like it was useful to have XML mean "XML only"

xml-project:
OK, missed that. Thanks for explaining.

PR #804, opened 2019-04-25
Fix 803

### LINKED ISSUE #803: Replace "XML media type" with "XML or HTML media type" in some places
I think I found some more places to change (some as a consequence of pr #799) :

1. XD0063 says it is an error if any inline content contains XML markup but has not an XML media type. I think here we need to add HTML media type as well.

2. In XD0054 we must add HTML media type

3. Section 16.10.1 should be renamed to "Inline XML and HTML content". In the first line we should add "HTML media type" additional to XML media type.

4. In section 10.2 under 1. we should add HTML media type. As a consequence we have to mention it in 2. too.

Can I have a go (or non-go) for this changes please.

xml-project:
Attempt to fix #803

### DIFF
--- a/xproc/src/main/xml/specification.xml
+++ b/xproc/src/main/xml/specification.xml
@@ -2172,7 +2172,8 @@ node, all the children of the document node are inserted.</para>
 
 <orderedlist>
 <listitem>
-<para>If the content type is an <glossterm>XML media type</glossterm>,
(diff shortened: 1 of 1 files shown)


xml-project:
@ndw thanks. Is there a problem with pr #802 - It is connected to this PR, so I would like to merge them together.

ndw:
The only problem is my inattention to detail. :-)

PR #806, opened 2019-04-29
Attempt to fix 805

### LINKED ISSUE #805: Serialization of XHTML needs a fix
In [3.1.2](http://spec.xproc.org/master/head/xproc/#html-documents) we say, define an HTML document as document with media type "text/html” and “application/xhtml+xml".
and then

> When an HTML document is serialized, it should be serialized using the HTML serializer (see [Serialization]) by default.

I think we should say that documents with "text/html" are serialized using method "html", but documents with "application/xhtml+xml" are serialized using method "xhtml".

Did I miss something?

xml-project:
Attempt to fix #805

### DIFF
--- a/xproc/src/main/xml/specification.xml
+++ b/xproc/src/main/xml/specification.xml
@@ -475,7 +475,7 @@ linkend="xml-serialization-31"/>) by default.
 Within XProc, they are <link linkend="xml-documents">XML documents</link>.
 HTML documents are identified by an HTML media type.
 <termdef xml:id="dt-HTML-media-type">The
-“<literal>text/html</literal>” and “<literal>application/xhtml+xml</literal>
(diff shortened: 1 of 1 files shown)


xproc/src/main/xml/specification.xml
```diff
@@ -475,7 +475,7 @@ linkend="xml-serialization-31"/>) by default.
 Within XProc, they are <link linkend="xml-documents">XML documents</link>.
 HTML documents are identified by an HTML media type.
 <termdef xml:id="dt-HTML-media-type">The
-“<literal>text/html</literal>” and “<literal>application/xhtml+xml</literal>
+  “<literal>text/html</literal>” and “<literal>application/xhtml+xml”</literal>
```
gimsieke: the closing quote after `application/xhtml+xml` should come after `</literal>`

PR #838, opened 2019-07-11
Speculative implementation of a p:param function

ndw:
Here's a first pass `p:param`. Unfortunately, I'm not confident that this is going to help.

The problem is that as a function, it can certainly get runtime values, but if the plan is to use this in the serialization options on `p:output`, then it has to be evaluated statically. You can't say that a function must be evaluated statically (well, you can, but I wouldn't).

At best, what we mean is that if the function appears in a context that is going to be evaluated statically, then it will be. But then we can't really, usefully speak of them as runtime values. They're compile time values or something and it's just fundamentally odd to have an ordinary function to get those.

Are there any cases **other** than serialization parameters where we feel we need to have user-specified values available statically?


xml-project:
I think to call the values "runtime values" is confusing because they must be available during static analysis. Places they are needed (may be I missed one, please add):

1. Serialization parameters
2. The select expressions on static variables.
3. Value templates in the attributes or descendants of p:input and p:output and map attributes on those descendants.
4. Expressions in use-when attributes used for conditional element exclusion.

My idea is that the function is available during static analysis. So the processor has to find a way to make this possible. To me (as you said) it makes only sense, if I can write this:

````
<p:store href="debug-line/info.xml" use-when="p:param('debug-mode')" />
````

xml-project:
@ndw How do you propose to go on? Merge the PR and then try to fix the open questions?
Since you broke functions into files with this PR, I cannot merge fix for #839 without breaking your PR.

ndw:
I'm wholly unsatisfied by the current attempt. Go ahead and break my PR, once I get the Balisage paper finished, I'll return my thoughts to this PR.

xml-project:
After thinking about it (and implementing it), I think we should do away with the function approach. I still believe that we could have a user set twin of p:system-property being useful during static analysis. 

The main reason for abandoning the idea is, that it leads to very ugly pipelines. The function-approach with the necessary default value means that I have to write the default value over and over again - just in case there is no user set value. The option approach here was much better: Set it once with the default value - use it anywhere.

But that still does not mean I am happy with the static option/static variable approach because of all the complexities it leads to. After all I am not sure that we need static options AND static variables. Static options seems to fit all needs (and can be set from the outside). Do we really need both?

Sorry for having just questions and not offering a solution. But may be it rings a bell for somebody.

ndw:
I agree. This is a disaster. I'll see if I can come up with another proposal.

PR #848, opened 2019-07-22
Fix #847 by making the error port primary

### LINKED ISSUE #847: Proposal: make the error port the default readable port in p:catch/p:finally
Observations (unless I’ve overlooked something, of course):

1. If you want to read the `error` port in a `p:catch`, you have to both name the catch and fully qualify the `p:pipe` (or `@pipe`) that reads from it.
```
...
<p:catch name=“catch”>
   <p:identity>
      <p:with-input pipe=“error@catch”/>
   </p:identity>
...
```
2. The default readable port, if there is one, comes from the step that precedes the `p:try`. That doesn’t seem like a very useful default.
3. Authors can always write pipes that point back to the preceding step if they want. And that will probably feel (relatively) natural as they’re reaching outside the confines of the container that they’re in.

It’s a little odd, I admit, but I still think it might be an improvement.

ndw:
I think this is the simplest resolution.

### DIFF
--- a/xproc/src/main/xml/specification.xml
+++ b/xproc/src/main/xml/specification.xml
@@ -4092,8 +4092,9 @@ modification:</para>
 
 <itemizedlist>
 <listitem>
-<para>The port named “<port>error</port>” on the <tag>p:catch</tag> is
-added to the <glossterm>readable ports</glossterm>.</para>
+<para>A primary input port named “<port>error</port>”
+is added to the <glossterm>readable ports</glossterm>
+on the <tag>p:catch</tag>.</para>
 </listitem>
 </itemizedlist>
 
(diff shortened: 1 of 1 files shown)


xml-project:
@ndw The following tests of the test suite need to be changed as a result:

- ab-drp-context-004
- ab-drp-context-014
- ab-drp-context-015
- ab-drp-context-023

ndw:
I created https://github.com/xproc/3.0-test-suite/issues/228 to track the test changes. Can't (reliably) check those until I change my implementation :-)

Feel free to fix them first if you get to them before I do.

xml-project:
@ndw Will do tomorrow.

PR #855, opened 2019-07-27
Fix #853

### LINKED ISSUE #853: Preceding or trailing whitespace in p:inline
In the [specs](http://spec.xproc.org/master/head/xproc/#inline-xml-content) for `p:inline` we say:

> Any preceding or following whitespace-only text nodes will be discarded.

I think the reference (preceding or following **what**) is ambiguous: Do we mean preceding/following the first none text node or do we mean before and after an element node (analogues to XML)?

ndw:
Clarify that whitespace is not trimmed within `p:inline`. Fix reference to `content-type` property in document properties. Clarify example.

### DIFF
--- a/xproc/src/main/xml/specification.xml
+++ b/xproc/src/main/xml/specification.xml
@@ -6108,12 +6108,11 @@ depends on the document's content type.
   <section xml:id="inline-xml-content">
     <title>Inline XML and HTML content</title>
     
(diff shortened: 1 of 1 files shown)


ndw:
Yes, I'm fine leaving this open until Gerrit or Erik comments.

gimsieke:
Fine with me

Issue #505, opened 2018-08-14
Scoping of (non-static) options


xml-project:
In XProc 1.0 the following pipeline has a static error, because `$option` is not visible in the XPath-expression:
````
<p:declare-step version="1.0" 
  xmlns:p="http://www.w3.org/ns/xproc"
  xmlns:mox="http://www.xml-project.com/morganaxproc">
  <p:output port="result" />
	<p:option name="option" select="'name'" />	
	<p:declare-step type="mox:test" >
		<p:output port="result" />
		<p:option name="inner" select="$option" />
		<p:add-attribute match="/*">
			<p:input port="source">
				<p:inline><doc /></p:inline>
			</p:input>
			<p:with-option name="attribute-name" select="$inner" />
                        <!-- HERE IS THE PROBLEM -->
			<p:with-option name="attribute-value" select="$option" />
		</p:add-attribute>
	</p:declare-step>
	<mox:test />
</p:declare-step>		
````
Now in the [XProc 3.0](http://spec.xproc.org/master/head/xproc/#environment) specs, we say:

> The environment of a step also includes the in-scope bindings for all of the variables and options “visible” from its lexical position.

I can't find any explanation of "visible" except the "shadowing" rule. But as `$option` in the marked position is not shadowed, one could assume it to be visible.
What did I miss? 
Should we be a little bit more explicit about visibility (or scoping) of options and variables?

gimsieke:
I, too, understand lexical scoping in such a way that embedded step declarations see options and static variables that are declared in the surrounding step declaration.

In https://github.com/xproc/3.0-specification/issues/826#issuecomment-507718221, @ndw asserts the following:

> […] Pipelines are self-contained. If I pick one up and drop it into a larger pipeline, it shouldn't suddenly break because of the static variables in the surrounding pipeline.

and in https://github.com/xproc/3.0-specification/issues/826#issuecomment-507343180, he argues

> I don't think we should make nested pipelines different from imported pipelines […]

So if we treat nested pipelines as if they are imported, we need to define “lexical scoping in XProc documents“ more precisely, because IMHO this contradicts everyone’s conception of lexical scope when they simply look at the XML document and its element nesting & order.

xml-project:
Sorry, could you give us (at least me) a hint, why you reopened this issue. 
As far as I understood our consensus "lexical scoping" means "lexical scoping inside the (embedded) step". 
This seems to be inline with the cited comments. Do you see this principle at risk?

[...]

PR #859, opened 2019-08-27
Trying to resolve #833

### LINKED ISSUE #833: Option default values
Consider this statement:

> Some steps accept options. The value of an option is the default value specified in its declaration, or a value provided by the caller of the step (overriding the default). If it has neither a default value nor a provided value, its value is the empty sequence.

I don't think it's that simple. Consider an option that has a type for which the empty sequence is not a valid value.

```
<p:option name="prefix" as="xs:NCName"/>
```

If I don't specify a value for `prefix` and the implicit default empty sequence value is used, then the value isn't legal according to the declared type.

Is that an error?

Is it an error if I specify `prefix="{()}"`?

Is it an error if I make the empty sequence default explicit?

```
<p:option name="prefix" as="xs:NCName" select="()"/>
```

If it's possible to come up with a rule that gives the same answer for all of these cases, I'd vastly prefer that over some context-dependent answer.

eriksiegel:
This was rather simple, but with the following note:

I mention "in-scope static variables".  I couldn't find in the core spec a specification of what the scope of a static variable was (from discussions: could be in an encompassing step, in a library, etc.). It's not defined yet. Since the jury on this is still out (#505) I think its safe to just mention it here as "in-scope", with the expectation we're going to write this down somewhere soon...

### DIFF
--- a/xproc/src/main/xml/specification.xml
+++ b/xproc/src/main/xml/specification.xml
@@ -2460,7 +2460,7 @@ Variable names are always expressed as
 
       <para>If a variable or option declares a type, the supplied value of the variable or option is converted to the
         required type, using the function conversion rules specified by XPath 3.1. <error code="D0036">It is a
-            <glossterm>dynamic error</glossterm> if the supplied value of a variable or option cannot be converted to
+            <glossterm>dynamic error</glossterm> if the supplied or defaulted value of a variable or option cannot be converted to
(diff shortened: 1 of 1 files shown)


eriksiegel:
That's it exactly: This fix references something we still have to make a final decision about, the scope of statics. I don't think that should postpone this tiny update

PR #861, opened 2019-08-29
Attempt to resolve #788

### LINKED ISSUE #788: Non-XML serialization
Currently we [say](http://spec.xproc.org/master/head/xproc/#xproc-non-xml-serialization):

> The names and values of non-XML serialization parameters are implementation-defined.

Doesn't it make sense to say, for JSON and text we align with
[XSLT and XQuery Serialization 3.1](https://www.w3.org/TR/xslt-xquery-serialization-31/)?

Did I miss something?

eriksiegel:
This is an attempt to resolve #788.

The whole prose for serialization was quite a historically grown jungle. I've tried to make it more clear but leave the important bits in. It does not handle the separate parameters any more (with the exception ot the method parameter, which is rather fundamental), it simply refers to [the serialization spec](https://www.w3.org/TR/xslt-xquery-serialization-31/).

Looking at the result I seem to have inadvertently changed indenting in other places. My apologies for that, that was not the intention but is now hard to roll back.

Anyway, This is one of the larger edits which are hard to judge without the final HTML at hand. Since I do not have any github pages setup, I will add my local copy of the HTML to the PR.

### DIFF
--- a/xproc/src/main/xml/specification.xml
+++ b/xproc/src/main/xml/specification.xml
@@ -1,96 +1,86 @@
 <?xml-model href="../../schema/dbspec.rng"
   type="application/xml"
   schematypens="http://relaxng.org/ns/structure/1.0" ?>
-<specification xmlns="http://docbook.org/ns/docbook"
-               xmlns:xlink="http://www.w3.org/1999/xlink"
-               xmlns:e="http://www.w3.org/1999/XSL/Spec/ElementSyntax"
-               xmlns:xi="http://www.w3.org/2001/XInclude"
-               xml:id='xproc'
-               class="ed"
-               version="5.0-extension w3c-xproc">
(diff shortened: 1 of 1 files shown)


eriksiegel:
[index.zip](https://github.com/xproc/3.0-specification/files/3555520/index.zip)

gimsieke:
Re source code indentation: If you only changed Sect. 16.3.1, you might want to create a new branch, only replace this particular section with the new text (keeping all other source code indentation intact) and then submit a new PR. We might run into unnecessary conflicts otherwise.

eriksiegel:
@gimsieke Yes, that's a good idea, I will do that.

eriksiegel:
I'm closing this one now, wil do it again with proper indentation later.

Issue #866, opened 2019-09-03
XProc defines a single, global namespace for static options.


xml-project:
I think this sentence should read:

XProc defines a single, global **scope** for static options.

ndw:
Okay. I agree that in the context of an XML language, "namespace" is confusing.

xml-project:
Fixed with #867

PR #867, opened 2019-09-03
Attempt to fix 865 and 866

### LINKED ISSUE #865: Wording of XS0091 still needs to be changed
After reading yesterdays PR in context, I think XS0091 needs a small update. Currently it reads:

> It is a static error if a p:option declared before the subpipeline begins shadows another option declared within the same p:declare-step.

The phrase "declared before the subpipeline begins" is a relict from static variables because afaik every option (static or not) is declared before the subpipeline begins. So the phrase is not wrong but may be misleading.

xml-project:
Attempt to fix #865 and #866

### DIFF
--- a/xproc/src/main/xml/specification.xml
+++ b/xproc/src/main/xml/specification.xml
@@ -2450,7 +2450,7 @@ Variable names are always expressed as
       <para>The values of static options are computed during <link linkend="initiating">static
           analysis</link>.</para>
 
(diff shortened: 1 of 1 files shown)


xml-project:
@gimsieke Thanks. Sure.

xml-project:
Waiting for @ndw to approve this before I merge.

PR #871, opened 2019-09-04
Attempt to fix 849

### LINKED ISSUE #849: Open questions on p:viewport
I think [`p:viewport`](http://spec.xproc.org/master/head/xproc/#p.viewport) needs to be updated to recent development:

> It is a compound step that processes a single **XML** document ....

I think we need to allow HTML documents also. And (if I am right with the next point) also Text documents.

> It is a dynamic error (err:XD0010) if the match expression on p:viewport does not match an element or document.

I think this is a leftover from XProc 1.0. Since we now use XDM i think `match` should be able to match anything except an attribute and a namespace.

> and the result that appears on the output port replaces the matched element.

I think we need to say more about what happens relative to the result's content-type.
Proposal:

1.  For an XML document and a HTML document, the matched element is replaced with all children of the result's document node.
2. For a Text document the matched element is replaced by the child of the result's document node. (This is the same as (1), but I listed it separately so can vote against it.
3. If it is a JSON document, it is an error.
4. If it is another document, it is an error.

xml-project:
Attempt to fix #849 

- Adapted p:viewport so it can process XML and HTML documents
- The match is no longer restricted to document-node or elements but can now select every node type (except attributes and namespace nodes).
- The result of the subpipeline can be an XML, HTML or text document.
- The content type changes, if match is "/" and the result is a text document.

### DIFF
--- a/xproc/src/main/xml/specification.xml
+++ b/xproc/src/main/xml/specification.xml
@@ -3589,9 +3589,9 @@ with the <link linkend="dt-standard-modifications">standard modifications</link>
 <title>p:viewport</title>
 
 <para>A viewport is specified by the <tag>p:viewport</tag> element. It
-is a <glossterm>compound step</glossterm> that processes a single XML
-document, applying its <glossterm>subpipeline</glossterm> to one or
-more subtrees of the document. </para>
+is a <glossterm>compound step</glossterm> that processes single XML or 
+HTML documents, applying its <glossterm>subpipeline</glossterm> to one or
+more subtrees of each document in turn.</para>
 
 <e:rng-pattern name="Viewport"/>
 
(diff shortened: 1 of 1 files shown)


xproc/src/main/xml/specification.xml
```diff
@@ -3652,21 +3653,34 @@ among the descendants of any matched node.</para>
...
-        element will effectively be deleted. If exactly one document appears, the contents of that
-        document will replace the matched element. If a sequence of documents appears, then the
+        node.</para>
+<para>If a document resulting from applying the subpipeline to the matched node is an XML document, 
+an HTML document, or a text document, all child nodes of the document node will be used to replace 
+the matched node. <error code="D0073">It is a <glossterm>dynamic error</glossterm> if the document 
+returned by applying the subpipeline to the matched node is not an XML document, an HTML document, 
+or a text document.</error>
+</para>
+<para>If no documents appears on the <port>output</port> port, the matched
```
gimsieke: If no documents appear (without “s”)

xml-project: Yap, thanks!

PR #872, opened 2019-09-04
Attempt to resolve #505

### LINKED ISSUE #505: Scoping of (non-static) options

eriksiegel:
Attempt to resolve #505 (clarify scoping of non-static option and variable names).

I also took the liberty of reorganizing section 14.2 a bit.

### DIFF
--- a/xproc/src/main/xml/specification.xml
+++ b/xproc/src/main/xml/specification.xml
@@ -2737,25 +2737,8 @@ xml:id="dt-visible">If two names are in the same scope, we say that
 they are <firstterm>visible</firstterm> to each other.
 </termdef></para>
 
-    <para>Six kinds of things are named in XProc:</para>
-    <orderedlist spacing="compact">
-      <listitem>
-        <simpara>Step types,</simpara>
-      </listitem>
-      <listitem>
-        <simpara>Steps,</simpara>
-      </listitem>
-      <listitem>
-        <simpara>Input ports,</simpara>
-      </listitem>
-      <listitem>
-        <simpara>Output ports,</simpara>
-      </listitem>
-      <listitem>
-        <simpara>Options and variables</simpara>
-      </listitem>
-    </orderedlist>
-
+<section>
+  <title>Scoping of step type names</title>
       <para>The scope of the names of the step types is the pipeline in which they are declared,
         including any declarations imported from libraries via <tag>p:import</tag>. Nested pipelines
         inherit the step types in scope for their parent.</para>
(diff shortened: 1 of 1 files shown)


xproc/src/main/xml/specification.xml
```diff
@@ -2812,31 +2795,52 @@ they are <firstterm>visible</firstterm> to each other.
...
-specified, its scope consists of the sibling elements that follow its
+</section>
+      
+<section>
+  <title>Scoping of non-static options and variables</title>
+<para>The scope of non-static option and variable names is determined by where
+they are declared. Their scope consists of the sibling elements that follow its
 declaration and the descendants of those siblings.
 </para>
+<para>Non-static options and variables declared in parent steps are not visible in child steps.</para>
```
gimsieke: Maybe make it a bit more precise: “parent step declarations” and “child step declarations”. If that’s what you mean. 

Otherwise one could ask oneself whether `$foo` isn’t visible in `p:for-each` in this scenario:

```xml
<p:viewport match="something" name="parent">
  <p:variable name="foo" select="/something/bar"/>
  <p:for-each name="child">
    <p:variable name="$baz" select="count($foo)"/>
    …
  </p:for-each>
</p:viewport>
```

The viewport is the parent step, and the for-each is the child step. `$foo` is clearly in scope in the child step. In order not to be misleading, you need to speak of parent and child declarations, I guess.

eriksiegel: Thanks, better still. Will change.

eriksiegel:
Sorry to ask, but anybody that now dares to approve this PR?

xproc/src/main/xml/specification.xml
```diff
@@ -2812,31 +2795,52 @@ they are <firstterm>visible</firstterm> to each other.
       <para><error code="S0036">All the step types in a pipeline or library <rfc2119>must</rfc2119>
           have unique names: it is a <glossterm>static error</glossterm> if any step type name is
           built-in and/or declared or defined more than once in the same scope.</error></para>
-
-<para>The scope of the names of the steps themselves is determined by
+</section>
+      
+<section>
+  <title>Scoping of step names</title>
+<para>The scope of the names of the steps (the values of the step's <code>name</code> attributes) is determined by
```
gimsieke: Please use typographic quotes (U+2019): step’s

eriksiegel: I changed the one you pointed out, but a quick search showed me a 's is used in 55 places. I suggest we do these kinds of finalizing editorial changes in somewhere near the end.

PR #882, opened 2019-10-03
Attempt to fix 880

### LINKED ISSUE #880: Document property "serialization"
In the [core specs](http://spec.xproc.org/master/head/xproc/#document-properties) we say, that it is 

> a dynamic error if a value is assigned to the serialization document property that is not of type map(xs:QName, item()*)

And in the [step specs](http://spec.xproc.org/master/head/steps/#c.set-properties) for `p:set-properties` we say nothing about this case.

1. I think we should handle both cases equally, so it is an error in both cases.
2. We should relax the rules for key "serialization" so that [QName-map magic](http://spec.xproc.org/master/head/xproc/#qname-handling) is applied here to. Therefor it is only an error if the processor can not convert the value for key "serialization" into a map with QName keys.

xml-project:
Attempt to fix #880 
Please mind there is a corresponding PR in the steps repo.

### DIFF
--- a/xproc/src/main/xml/specification.xml
+++ b/xproc/src/main/xml/specification.xml
@@ -404,7 +404,7 @@ XProc pipeline as a map (<code>map(xs:QName, item()*)</code>).</termdef></para>
               properties for the document. If present, it's value <rfc2119>must</rfc2119> be of type
                 <code>map(xs:QName, item()*)</code>. <error code="D0070">It is a <glossterm>dynamic
                   error</glossterm> if a value is assigned to the <code>serialization</code>
-                document property can not be converted into <code>map(xs:QName, item()*)</code> according
+                document property cannot be converted into <code>map(xs:QName, item()*)</code> according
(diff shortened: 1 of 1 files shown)


xproc/src/main/xml/specification.xml
```diff
@@ -404,7 +404,8 @@ XProc pipeline as a map (<code>map(xs:QName, item()*)</code>).</termdef></para>
               properties for the document. If present, it's value <rfc2119>must</rfc2119> be of type
                 <code>map(xs:QName, item()*)</code>. <error code="D0070">It is a <glossterm>dynamic
                   error</glossterm> if a value is assigned to the <code>serialization</code>
-                document property that is not of type <code>map(xs:QName, item()*)</code>.</error>
+                document property can not be converted into <code>map(xs:QName, item()*)</code> according
```
gimsieke: … document property can not be converted …
→
… document property that cannot be converted …

Otherwise approve

Issue #889, opened 2019-10-19
Be more explicit about syncing base-uri document properties with base URIs


gimsieke:
In the [Document Properties](http://spec.xproc.org/master/head/xproc/#document-properties) section we say:

> The processor is responsible for assuring that the `base-uri` property matches the base URI of each document produced on every output port.

I think it is obvious that this implies: If, for example, `p:xslt` produces documents with base URIs on its secondary output port, the processor needs to equip these documents with a corresponding `base-uri` property.

But is it also implied that if a property is set using [`p:set-properties`](http://spec.xproc.org/master/head/steps/#c.set-properties) or a `document-properties` option (for example on `p:inline`), what is then returned by `base-uri(/)` should reflect the new value?

I think it is implied since otherwise the document property and the base URI wouldn’t match. Or if the base URI weren’t affected but property and base URI were still supposed to match, `p:set-properties` couldn’t be used to alter the base URI.

Therefore I think also setting the property needs to manipulate the base URI, and I think we should make this explicit. 

And we probably need to say which has precedence for `p:inline`, `p:load`, `p:document` and maybe some more: 
* The `base-uri` property that comes from the `properties`/`document-properties` map 
* or the base URI of the `p:inline` / the loaded document?

I’d say the map needs to have precedence because otherwise you cannot override a base URI of a document node.

xml-project:
I agree with the general line of your thoughts. However I have two remarks

- Just before the phrase you quoted, the specs say: "If no such key is present, the document has no base URI." Unless I am totally missunderstanding this phrase, this rules out the p:xslt case you made, unless pipeline authors explicitly use set-property. Here the two sentences obviously contradict each other. I think we should remove the sentence I quoted, which is inline with your approach.
- My second point results from the divergence of base uri (as defined in the [XDM specs](https://www.w3.org/TR/xpath-datamodel-31/#dm-base-uri)) and xs:anyURI (as defined in [XMLSchema](https://www.w3.org/TR/xmlschema11-2/#anyURI)) A base uri is defined with refence to "URI reference" (RFC 3986). So we have either to invent a mechanism converting the base-uri property (xs:anyURI) to the document's base uri (URI). Or we have to say, that the base-uri property has to be a URI. Do you see any other, hopefully easier and less intrusive, solution?

xml-project:
[...]

PR #890, opened 2019-10-20
Shot at 889

xml-project:
First shot at #889 (only core specs). Changes on p:load and p:set-properties (step specs) are pending.

Does this work to clarifiy the relation and to solve the synch problem?

### DIFF
--- a/xproc/src/main/xml/specification.xml
+++ b/xproc/src/main/xml/specification.xml
@@ -390,9 +390,9 @@ XProc pipeline as a map (<code>map(xs:QName, item()*)</code>).</termdef></para>
           </term>
           <listitem>
             <para>The value of the (optional) “<code>base-uri</code>” property identifies the base URI of the
-              document. If no such key is present, the document has no base URI. The processor is
-              responsible for assuring that the <code>base-uri</code> property matches the base URI
-              of each document produced on every output port.</para>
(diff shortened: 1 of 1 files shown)


xml-project:
Let's see what @ndw and @eriksiegel have to say.

@gimsieke 
> In the core spec, there needs to be additional clarity what the base-uri key of p:document/@document-properties does to the base URI of the read document (it should set the base URI to the value associtated with the base-uri key).

As we decided to define p:document completely via p:load, I think the clarification there should also contribute to the case you mentioned.

xproc/src/main/xml/specification.xml
```diff
@@ -390,9 +390,9 @@ XProc pipeline as a map (<code>map(xs:QName, item()*)</code>).</termdef></para>
           </term>
           <listitem>
             <para>The value of the (optional) “<code>base-uri</code>” property identifies the base URI of the
-              document. If no such key is present, the document has no base URI. The processor is
-              responsible for assuring that the <code>base-uri</code> property matches the base URI
-              of each document produced on every output port.</para>
+              document. For XML document, HTML documents, and text documents, the value of “<code>base-uri</code>”
```
ndw: "For XML document*s*, HTML ..."

ndw:
Looks ok to me.

xml-project:
Will supply the second half (p:load / p:set-properties) this afternoon.

PR #891, opened 2019-10-22
Added sentence as requested

### LINKED ISSUE #889: Be more explicit about syncing base-uri document properties with base URIs

xml-project:
Fix #889

### DIFF
--- a/xproc/src/main/xml/specification.xml
+++ b/xproc/src/main/xml/specification.xml
@@ -2862,6 +2862,8 @@ attribute. The <tag class="attribute">xml:base</tag> attribute
 <rfc2119>may</rfc2119> appear on any element in a pipeline and has the
 semantics outlined in <biblioref linkend="xml-base"/>.</para>
 
+<para>The base URI of a document node may be manipulated by changing the base-uri 
+ document property.</para>
 </section>
     <section xml:id="xml-id-attribute">
       <title>Unique identifiers</title>


xml-project:
I just did what I was told.

Issue #892, opened 2019-10-23
Nasty consequence of manipulating base uris


xml-project:
After implementing our new ruling yesterday evening, the following problem of allowing to manipulate document base URIs came up:

Until yesterday we had either document with a proper base URI (as a consequence of p:inline, p:load, p:xslt etc) and documents without a base URI, e.g. the result from p:count or p:compare.

Now since we allow pipeline authors to overwrite proper base URI with arbitrary string in the `document-properties` map, an author might attempt to remove document's base URI or overwrite the proper value with a value which is not a proper URI.

Since neither Saxon nor XOM accept an invalid URI as value for the document's base URI, and a number of steps rely on a valid URI to resolve relative URIs, I propose to have a new dynamic error which is raised, when on `p:inline`, `p:load` and `p:set-property` whenever a 'base-uri' key is present but not associated with a valid URI.

gimsieke:
You can’t completely prevent users from shooting themselves into their feet. If someone supplies a malformed URI for `p:load` or `p:store`, they will get a dynamic XProc error or even a Java exception such as “URI is not hierarchical”.

I agree to some extent, though. But as I said on other occasions, we shouldn’t enforce strict RFC 3986 compliance. Unescaped UTF-8 characters, `file:/C:/` instead of `file:///C:/`, etc., often resolve on the given OS platform. These URIs should only raise an error if they don’t resolve to files, not if they don’t strictly comply with RFC 3986.


So I 

Currently, if you use a malformed URI, you will get a Java exception, for ex. .

xml-project:
>But as I said on other occasions, we shouldn’t enforce strict RFC 3986 compliance. Unescaped UTF-8 characters, file:/C:/ instead of file:///C:/, etc., often resolve on the given OS platform. These URIs should only raise an error if they don’t resolve to files, not if they don’t strictly comply with RFC 3986.

Sorry, but this is not an option here because we are not talking about resolving files, but setting the base uri property per [XML base](https://www.w3.org/TR/xmlbase/). And here a proper URI is required. If you look at the APIs there is no way to supply your examples as base URI for Saxon or XOM. Both require instances of java.net.uri. 

The only alternative I see currently is to revert everything I have done yesterday and forbid manipulation of the base URI property.

gimsieke:
I created a document containing an U+2192 UTF-8 char
```xml
<doc xml:base="file:/C:/test→UTF-8.xml"/>
```
and transformed it using Saxon 9.8.0.15
[...]

PR #895, opened 2019-10-24
Revert PR 890

xml-project:
Attempt to solve #892 and https://github.com/xproc/3.0-steps/issues/251
by reverting pr #890 

As the discussion has shown, this PR was premature.

### DIFF
--- a/xproc/src/main/xml/specification.xml
+++ b/xproc/src/main/xml/specification.xml
@@ -390,9 +390,9 @@ XProc pipeline as a map (<code>map(xs:QName, item()*)</code>).</termdef></para>
           </term>
           <listitem>
             <para>The value of the (optional) “<code>base-uri</code>” property identifies the base URI of the
-              document. For XML documents, HTML documents, and text documents, the value of “<code>base-uri</code>”
-              is the base URI property of the document node. For other document types it is a property 
-              the processor keeps track of. If no such key is present, the document has no base URI.</para>
+              document. If no such key is present, the document has no base URI. The processor is
+              responsible for assuring that the <code>base-uri</code> property matches the base URI
+              of each document produced on every output port.</para>
           </listitem>
         </varlistentry>
         <varlistentry>
@@ -5896,10 +5896,7 @@ the <tag class="attribute">document-properties</tag> map contains a
 from the statically determined content type.</error></para>
 
 <para>The base URI of the document is the base URI of the
-<tag>p:inline</tag> element or of the parent element in
-the case of an implicit inline. If <tag class="attribute">document-properties</tag>
-provides a value for “<literal>base-uri</literal>”, this value is the
-base URI of the document.</para>
+<tag>p:inline</tag> element.</para>
 
 <para>How the content of a <tag>p:inline</tag>
 element is interpreted depends on the document's content type and the


xml-project:
I think we already have a way to change the base URI property of a document node:
```
<p:identity>
  <p:with-input>
    <p:inline xml:base="http://new-uri">{.}</p:inline>
  </p:with-input>
</p:identity>
```
If you look at all the problems and additional requirement that are to be fulfilled as a result of PR #890 I think it is simple not worth it.

gimsieke:
In #892 you say:

> Now since we allow pipeline authors to overwrite proper base URI with arbitrary string in the document-properties map, an author might attempt to remove document's base URI or overwrite the proper value with a value which is not a proper URI.

Can’t the `p:inline` solution that you propose above be used by pipeline authors to assign “improper” base URIs to documents? If you want to prevent authors from doing so, you need to say that a base URI may never be changed, or only changed to what is a “valid” URI. 
In any case, I don’t see why, on the one hand, you want to allow the `p:inline` solution (or my clumsy `p:xslt` solution that already works with 1.0), in which authors can shoot themselves in their feet ad libitum, while on the other hand you want to rule out manipulation of a document node’s base URI by changing the `base-uri` document property.

xml-project:
> Can’t the p:inline solution that you propose above be used by pipeline authors to assign “improper” base URIs to documents?

No, because it is an error according to the definition of "[xml:base](https://www.w3.org/TR/xmlbase/)" : "The terms base URI and relative URI are used in this specification as they are defined in [RFC 3986]."

gimsieke:
So your main issue is that `base-uri` is `xs:anyURI` that can virtually be any string, according to XML Schema 1.1 Part?

Can we agree and specify that, in principle, the same (stricter, that is, RFC 3986 or rather LEIRI) rules apply to the `base-uri` property that apply to `@xml:base` attributes? 

But, as [XDM 3.1](https://www.w3.org/TR/xpath-datamodel-31/#namespace-names) says about namespace names, “implementations **may** reject character strings that are not valid URIs or IRIs, but they are not _required_ to do so.” 

Or phrased differently: “Although the `base-uri` property is an XDM 3.1 / XSD 1.1 `xs:anyURI` value, implementations may, but are not required to, reject strings that are not valid URIs according to the `xml:base` specification. Likewise, they may accept strings for the `@xml:base` attribute that are castable to `xs:anyURI` but do not strictly comply with the `xml:base` specification. It is a dynamic error XD… if the `base-uri` property or the `xml:base` attribute has a value that may not be used as a base URI.”

In practice, for example Saxon accepts `@xml:base` values that do not conform to the strict spec. This may or may not be a problem, and an XProc implementation is free to apply additional checks and raise this error.

xml-project:
> So your main issue is that base-uri is xs:anyURI that can virtually be any string, according to XML Schema 1.1 Part?

The short answer would be "no". The longer one: I did not mention xs:anyURI in the issues and there is also no reference to xs:anyURI in [xml:base] or [RFC 3986]. Therefor I do not understand how this comes into play here.

gimsieke:
I only try to understand your point. On multiple occasions you said that I didn’t get the point. Can you try to explain again what is problematic about the PRs that you want to revert?

xml-project:
@gimsieke I will try to send you a mail over the weekend, but I am not sure this will work out, because I am busy with other things. Or we wait, until Norm tries to implement the specs and discovers the problems. Or, you could simply trust me. ;-))

gimsieke:
Or you can try to explain it to me in a phone call. 

Anyway, Cologne is only a fortnight away, and we can discuss it there. 

Most of the time when there is a lack of common understanding of some matter, discussing it face to face is more productive than discussing it in a conference call which in turn is more productive than discussing it on some forum on the Internet…

Be assured that the reason that I didn’t quite grasp your point yet it is my dumbness and not bad faith / passive-agressiveness ;)

xml-project:
I guess it may be simply my lack of expressiveness in English.

gimsieke:
Ok, it was good to discuss it on the phone, @xml-project. Will send the plan how we propose to tackle the base URI issue to the editors later today.

PR #896, opened 2019-11-04
Issue889

gimsieke:


gimsieke:
I think it doesn’t build because I used an `example` element and some magic seems to be associated with them. I’ll try to turn it into a note.

xml-project:
Actually I think the problem is, what Travis says it is: You used "xml:space" at a place, where it is not allowed in DocBook. Since I am not an expert, I have to trust Travis there. If you think, it is allowed, you know where to complain.

gimsieke:
Superseded by #918

PR #903, opened 2019-11-09
Minor fixes

ndw:
Small fixes that came out of today's meeting


xproc/src/main/xml/specification.xml
```diff
@@ -5896,7 +5899,10 @@ the <tag class="attribute">document-properties</tag> map contains a
 from the statically determined content type.</error></para>
 
 <para>The base URI of the document is the base URI of the
-<tag>p:inline</tag> element.</para>
+<tag>p:inline</tag> element or of the parent element in
+the case of an implicit inline. If <tag class="attribute">document-properties</tag>
```
gimsieke: role="attribute"?

ndw: Nope. @class.

PR #913, opened 2019-11-17
Define the scope of static option names

### LINKED ISSUE #868: Still struggling to grasp the single scope rule for static options
Consider:
````
<p:declare-step>
  <p:declare-step type="step:one">
    <p:option name="opt" static="true" select="42" />
    <!-- ... -->
  </p:declare-step>
  <p:declare-step type="step:two">
    <p:variable name="var" select="$opt" /> <!-- (1) -->
  </p:declare-step>
  <p:variable name="var" select="$opt" />   <!-- (2) -->
````
The single scope rule first says that I am not allowed to declare another option/variable with the same name in the global scope.

But what about visibility: If there is **one** scope, it seems to imply to me, that (1) and (2) are allowed. In other words, I can use any static option in an XPath expression, provided the declaration of this option lexically precedes the use.

What did I miss? And if I did not miss a think, should we state this more explictly in the specs?

ndw:
Fix #868

### DIFF
--- a/xproc/src/main/xml/specification.xml
+++ b/xproc/src/main/xml/specification.xml
@@ -2791,12 +2791,51 @@ declaration and the descendants of those siblings.
 </section>
       
 <section>
-  <title>Scoping of static options</title>
+  <title>Scoping of static option names</title>
 
-  <note  role="editorial">
-    <title>TBD</title>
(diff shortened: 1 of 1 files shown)


ndw:
Good point. We don't mention that about step types either. Will fix.

PR #930, opened 2019-12-25
Fixed typo

xml-project:
Added a missing closing hypen. Don't think that is controversial.


xml-project:
Uups.

xml-project:
I think you should like it now. 😇

PR #944, opened 2020-02-20
for-each example

### LINKED ISSUE #943: p:for-each example uses pipe="chapters@current"
The section 15.2.2 in https://spec.xproc.org/master/head/xproc/#p.for-each has an example 5 "A Sample For-Each" which declares `<p:for-each name="chapters">` and then inside uses `<p:with-input port="source" pipe="chapters@current"/>`.

Isn't that supposed to be `<p:with-input port="source" pipe="current@chapters"/>`, as the syntax wants the port name first and then the step name after the `@` sign?

gimsieke:
replace `chapters@current` with `current@chapters`

fix #943

### DIFF
--- a/xproc/src/main/examples/for-each.xml
+++ b/xproc/src/main/examples/for-each.xml
@@ -17,7 +17,7 @@
   </p:xslt>
 
   <p:xslt name="make-fo">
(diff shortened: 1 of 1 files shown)


gimsieke:
I didn’t consider this change substantive.

PR #977, opened 2020-06-05
Attempt to fix 975

### LINKED ISSUE #975: Serialization parameters
Concerning the serialization parameters we say in [16.3.1.1. Serialization method ](https://spec.xproc.org/master/head/xproc/#serialization-method)(last bullet point):

> Parameter omit-xml-declaration is set to true.

In the very next section the third bullet point reads:

> The omit-xml-declaration must be supported. If the value is not specified or has the value no, an XML declaration must be produced.

I am not sure I understand, why this is not contradictory.

xml-project:
Attempt to fix #975 as proposed.
@Gertone's point about a warning is still open because I am not sure whether this is going to be a "must/should/may" and how to incorporate the concept of a warning.
May be we should discuss this in next editors call.

### DIFF
--- a/xproc/src/main/xml/specification.xml
+++ b/xproc/src/main/xml/specification.xml
@@ -5070,7 +5070,7 @@ document</emphasis> and provide it on the output port. It
         </listitem>
         <listitem>
           <para>Parameter <literal>omit-xml-declaration</literal> is set to
-            <literal>true</literal>.</para>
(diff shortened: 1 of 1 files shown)


xml-project:
Please don't approave this PR because no Travis check has been performed.

xml-project:
Triggered Travis by hand, now this PR if Travis checked.

PR #984, opened 2020-07-02
Fix vt

xml-project:
Hi @ndw 
I tried to fix the text on VTs as discussed. Could you please check that I got are results right.


ndw:
I proposed glossterm markup around "default readable port" but if you don't think that's necessary, I'm fine with it.

Issue #983, opened 2020-06-29
Allow @xml:space on <p:inline>?


xatapult:
Maybe too late (feature freeze) but I dearly miss the ability to specify `xml:space="preserve"` on `p:inline` to prevent pretty printing of the contents by an editor.

Any chance we might still add this or is it really too late? I think it would just be a schema update, no changes to processors required.

xatapult:
Or add as a general attribute, just like `xml:id`?

gimsieke:
`@xml:space`, `@xml:lang`, `@xml:base`, and `@xml:space` should be permitted on (almost?) every element IMHO.

ndw:
Yes, absolutely.

xml-project:
FEATURE FREEZE!

xatapult:
Ah, clash of opinions. Let's discuss Wednesday.

xatapult:
I'm not sure that allowing `xml:base` is *just* a schema change. That has semantics attached.

gimsieke:
It’s already allowed where `@xml:id` is allowed.
https://github.com/xproc/3.0-grammar/blob/master/src/main/schema/core30.rnc#L177

xml-project:
> Let's discuss Wednesday.

But it is not Wednesday, but the 8th, isn't it.

Just another delay! With this strategy I am not sure we will finish the core specs before the end of this millenium!

xml-project:
> I'm not sure that allowing xml:base is just a schema change. 

xml:id and xml:base are allowed on any element says the [spec](https://spec.xproc.org/master/head/xproc/#common-attr).

xml-project:
@gimsieke Sorry, overlapping answers.

xatapult:
Sorry *next* week Wednesday, July 8th, 17:00.

xml-project:
Proposal: Defer to XProc 3.1

xml-project:
@gimsieke 
> It’s not a feature that creates new work items for any implementor I guess

Of course that is a big change for the implementors, because current those attribute are forbidden, therefor an error is raised.

xml-project:
Fixed with #985 and fix in grammar repo.

PR #985, opened 2020-07-03
Listing more xml attributes

xml-project:
As the spec already says "Attributes from the XML namespace are allowed anywhere. " calling out `xml:lang` and `xml:space` does not seem necessary, but it also does no harm. 
The key problem seems to be not the specs, but the grammar which does not reflect the prose in a correct way.
issue #983


gimsieke:
In my opinion we don’t need to say this in the change log if these attributes have been allowed by the spec all the time. (Thanks for spotting this, Achim!)

xml-project:
@gimsieke I am confused because you did not approve this PR. Does this mean you want to call out xml:id and xml:base in the specs, but not xml:lang and xml:space?

gimsieke:
Sorry, I should have uncollapsed the context that Github hid. I just assumed these listitems were in the change log.

PR #1000, opened 2021-01-13
Editorial proposal: Remove left-over para

xml-project:
I think the deleted para is a left-over from the times there were 'p:pipeline' and 'p:declare-step'.


xatapult (approved):

xatapult:
Obvious. Thanks for spotting this.

PR #1009, opened 2021-04-12
Attempt to clarify conversion of option shortcut values

ndw:
This PR attempts to resolve an issue raised in an email conversation where the rules for AVTs in option shortcuts appeared to contradict the rules for AVTs in attribute values generally.


xml-project (commented):
I would prefer *function conversion rules* instead of *normal conversion rules*. I think this is essentially what you mean, but the first term is used in the XPath specs, so it is easier to find.

ndw:
Yes, thank you, that is better. I pushed a fix.

gimsieke (approved):

Issue #995, opened 2020-09-03
Clarify whether a dependency is established between two steps if the following step could possibly use a port of the preceding as DRP


gimsieke:
see the discussion on [Twitter](https://twitter.com/gimsieke/status/1301540216688214016)

The paragraph above [this anchor](https://spec.xproc.org/master/head/xproc/#err.inline.D0001.5) reads: 

> If collection is true, the context item for the expression is undefined. Otherwise, the context item for the expression comes from the document connections, if they are specified. If they are not specified, the context item comes from the default readable port of the step. If no default readable port exists, the context item is undefined.

I understand Achim [as asserting](https://twitter.com/xml_project/status/1301535240607891456) that a dependency will only be established if an option actually refers to the context item.

I wonder which is true. Maybe we should state it more explicitly?

xml-project:
> I understand Achim as asserting that a dependency will only be established if an option actually refers to the context item.

Right. But it is not only true for construction with `@collection=true`, but applies to all XPath expression.

Otherwise the whole section on [parallelism](https://spec.xproc.org/lastcall-2020-08/head/xproc/#parallelism) makes no sense.

> Maybe we should state it more explicitly?

I agree: Having parallelism of unconnected steps speeds up pipeline execution a lot. Making potentially connected steps actually connected would take away almost any parallelism. Since all almost all steps have primary output port and almost all steps have options (which **could** use the primary output port as context item), almost all pipelines would form a straight line of execution.

xml-project:
@ndw 

1.  I do not think that a processor is required to do so, but it may.
2. To my understanding the order of execution is determined by the flow of documents. If you change `count(/*)`to `3+4` you change the flow because you take away the consumer.
3. The concept of "order of execution" (as opposite from the flow of documents) to my reading only comes into play when we look at side-effects as interacting with the file system.
4. Doing away almost any chance of having parallelism is not especially appealing to me.

xml-project:
On a second thought: The problem @dmj raised with
````
<p:file-delete />
<p:file-mkdir />
````
is not totally connected to the question of context item. Suppose:
````
<p:file-mkdir href="apath" />
````
There is just a string in `href`, so there is no context-item. But if I write
````
<p:file-mkdir>
  <p:with-option name="href" select="'apath'" />
</p:file-mkdir>
````
there is one. So in the first case it is not guaranted that mkdir is executed after deleted, but in the second case it would be guaranteed. That seems strange to me (to say the least).

[...]

PR #1024, opened 2021-07-22
First attempt to resolve issue 995

ndw:
This is my first attempt to clarify the binding of the default readable port per #995. At the moment, I've only added some prose to the "Connections" section.

### DIFF
--- a/xproc/src/main/xml/specification.xml
+++ b/xproc/src/main/xml/specification.xml
@@ -1163,6 +1163,52 @@ value template in a <tag>p:inline</tag>.
             possibilities for this connection is exactly the same set as for any other input port
             within the current <glossterm>environment</glossterm>.</para>
 
+   <section xml:id="connecting-the-drp">
+<title>Connections and the Default Readable Port</title>
+
+<para>The <glossterm>default readable port</glossterm> is a
+convenience for pipeline authors. In the pipeline document which
+describes a pipeline, steps are often represented with sequential
+elements and it is very common for the output of one step to form the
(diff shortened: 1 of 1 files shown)


gimsieke (commented):
This is fine, but I think we might add another example or note. A scenario as @dmj described it in https://twitter.com/_dmaus/status/1301537247079079938. If you have a pipeline that copies file A to B and then B to C, it might throw an error because it attempts to execute the B→C operation before the A→B operation has finished or even started. `p:file-copy` has a primary output but the next copy operation’s `@href` and `@target` attributes are just string constants, they don’t depend on the previous step’s output. Without explicit dependency declaration, the steps might be executed in parallel or in “wrong” order.

ndw:
Quite right. I added those examples and made some clarifying editorial changes.

xproc/src/main/xml/specification.xml
```diff
@@ -1179,12 +1179,13 @@ sibling. Consider the following fragment:</para>
                  attribute-value="{count(/*/*)}"/>]]></programlisting>
 
 <para>The output from the add XML base step is the natural input to the
-add-attribute step. That’s the source for both the document that will
-be processed by the step and the document that will be used as the
+add-attribute step. The add XML step is the source for both the document that will
```
gimsieke, line 1182: add XML step → add XML base step

gimsieke (approved):

xproc/src/main/xml/specification.xml
ndw: 👍

PR #1026, opened 2021-07-24
Rework "handling QNames" as "implicit casting"

### LINKED ISSUE #1001: Question about AVTs and QName-magic
Not sure this is right, so pointing out my mistake to me would help.
Consider:
````
<p:wrap-sequence wrapper="wrapper" />
````

[Section 10.1.](https://spec.xproc.org/master/head/xproc/#attribute-value-templates) says:

> The value of an attribute that contains attribute value templates is a single string (the concatenation of the string values of the evaluated templates and non-template parts) as an xs:untypedAtomic.

So: The value of $wrapper is xs:untypedAtomic. Right?

But the option `wrapper`on `p:wrap-sequence` is declared as `xs:QName` so this is the place, where "[QName handling](https://spec.xproc.org/master/head/xproc/#qname-handling)" has to come in. But QName handling is only defined for xs:string -> xs:QName, not for xs:untypedAtomic -> xs:QName.

Therefor the above snippet `p:wrap-sequence` has to raise an error [XD0068](https://spec.xproc.org/master/head/xproc/#err.D0068). Right?

Could somebody please tell we, why I am wrong? What did I miss?

### LINKED ISSUE #1012: Error in example?
In section [7.2.2. XPath in XProc](https://spec.xproc.org/lastcall-2020-08/head/xproc/#xpath-context) the code example uses:
````
<p:load name="read-from-home">
  <p:with-option name="href" select="concat($home,'/document.xml')"/>
</p:load>
````
I would argue that this should fail with XD0036 because `select` returns an xs:string, but `href` requires an xs:anyURI. As function conversion rules apply here, the xs:string cannot be converted to xs:anyURI. Hence the error.
The example is a verbatim copy of the one used in XProc 1.0 specification, so I guess that is the reason for this problem.
Did I miss something?

### PR DESCRIPTION
**Pull request by ndw**
This PR rewords the section about implicit casting of QNames so that it also covers implicit casting of `xs:anyURI`s. I believe it will fix #1001 and fix #1012.

### DIFF
--- a/xproc/src/main/xml/specification.xml
+++ b/xproc/src/main/xml/specification.xml
@@ -485,7 +485,7 @@ document has no base URI.</para>
                 <code>map(xs:QName, item()*)</code>. <error code="D0070">It is a <glossterm>dynamic
                   error</glossterm> if a value is assigned to the <code>serialization</code>
                 document property that cannot be converted into <code>map(xs:QName, item()*)</code> according
-                  to the rules in <link linkend="qname-handling">QName handling</link>.</error>
+                  to the rules in <link linkend="implicit-casting">Implicit Casting</link>.</error>
               Serialization properties control XML serialization as defined by <biblioref
                 linkend="xml-serialization-31"/>. See also <xref linkend="serialization"/>. </para>
             <para>Some steps, like <code>p:xslt</code> and <code>p:xquery</code>, can specify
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: xproc/src/main/xml/specification.xml [hunk 1]
```diff
@@ -2692,8 +2692,26 @@ Variable names are always expressed as
...
+      <para>For the most part, the rules for casting between types in XPath work the way authors expect.
+      You can type “3” for a decimal, you don’t have to type “3.0”, and an untyped atomic value, for example in
+      an attribute such as <code>limit="3"</code>, is implicitly cast to a number if that’s the required type.
+      You don’t have to type <code>limit="{xs:int(3)}"</code>.</para>
+
+      <para>Unfortunately, there are two common cases in XProc
+      that are not handled by the XPath conversion rules: conversions
+      from strings to QNames or URIs. (Technically, conversions
+      from <type>xs:untypedAtomic</type> or <type>xs:string</type> (or from types
+      derived from <type>xs:string</type>) values to
```
gimsieke, line 2707: suggestion: replace the inner nested parentheses with commas

xproc/src/main/xml/specification.xml
```diff
@@ -2740,22 +2760,49 @@ Variable names are always expressed as
...
+        <type>xs:anyURI</type> are processed specially. The value (or values) are converted to
+        <type>xs:anyURI</type>s:</para>
+
+      <orderedlist>
+        <listitem>
+          <para>If the value supplied for the option is an instance of <type>xs:anyURI</type> then that value is used.
+          </para>
+        </listitem>
+        <listitem>
+          <para>If the value supplied for the option is an instance of <type>xs;untypedAtomic</type> or
```
gimsieke, line 2791: `xs;untypedAtomic` → `xs:untypedAtomic`

gimsieke (approved):
with minor corrections/suggestions

PR #1032, opened 2021-09-19
Clarify that default bindings are resolved dynamically

### LINKED ISSUE #1031: Are default input bindings resolve statically or dynamically?
Consider:
```
<p:declare-step type="ex:step">
  <p:input port="source">
    <p:document href="doc.xml"/>
  </p:input>
  ...
</p:declare-step>
```
Is `doc.xml` resolved and retrieved at "compile time", meaning the default will be the same for every use of the pipeline, or is the default the document reference itself which must be resolved at "run time" when the pipeline is being prepared for execution?

I think it should probably be resolved statically, but the spec doesn't say.

ndw:
Fix #1031

### DIFF
--- a/xproc/src/main/xml/specification.xml
+++ b/xproc/src/main/xml/specification.xml
@@ -4873,20 +4873,20 @@ author to exclude some namespace declarations in inline content, see <tag>p:inli
 </varlistentry>
 </variablelist>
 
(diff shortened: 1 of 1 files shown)


xatapult (approved):
Sorry, missed it.

PR #1034, opened 2021-10-11
The p:urify() function redrafted

ndw:
This is my attempt to rewrite the `p:urify()` function in what I hope is a faithful interpretation of the intent. It is backed up by 288 test cases in [a PR](https://github.com/xproc/3.0-test-suite/pull/539) on the test suite.

(The tests are also available in a slightly more compact form in my implementation:

* https://github.com/ndw/xmlcalabash2/blob/master/src/test/scala/com/xmlcalabash/test/UrifyWindowsSpec.scala
* https://github.com/ndw/xmlcalabash2/blob/master/src/test/scala/com/xmlcalabash/test/UrifyNonWindowsSpec.scala

Those are in a Scala testing DSL, but I don't think they're too difficult to understand.)

To review the changes in a formatted spec, see https://xpspectest.nwalsh.com/nw-urify/head/xproc/#f.urify

### DIFF
--- a/xproc/src/main/xml/specification.xml
+++ b/xproc/src/main/xml/specification.xml
@@ -1921,286 +1921,565 @@ interpreted as follows:</para>
 normally during static analysis.</para>
 </section>
   
-    <section xml:id="f.urify">
-      <title>Transform file system paths into URIs and normalize URIs</title>
-      <para><function>p:urify</function> is a function that attempts to transform file system paths
-        into file URIs (<biblioref linkend="rfc3986"/>). If a presumptive yet not fully compliant
-        URI is given as an argument, <function>p:urify</function> will perform operations such as
-        percent-encoding and path normalizations in order to make it compliant.</para>
+<section xml:id="f.urify">
+<title>Transform file system paths into URIs and normalize URIs</title>
 
-      <methodsynopsis>
-        <type>xs:string</type>
-        <methodname>p:urify</methodname>
-        <methodparam>
-          <type>xs:string</type>
-          <parameter>filepath</parameter>
-        </methodparam>
-        <methodparam>
-          <type>xs:string?</type>
-          <parameter>basedir</parameter>
-        </methodparam>
-      </methodsynopsis>
+<para>The <function>p:urify</function> function attempts to
+transform file system paths into file URIs (<biblioref
+linkend="rfc3986"/>). If a presumptive yet not fully compliant URI is
+given as an argument, <function>p:urify</function> employs a series
+of heuristics to resolve the string into a URI.</para>
 
-      <methodsynopsis>
-        <type>xs:string</type>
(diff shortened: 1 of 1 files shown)


xproc/src/main/xml/specification.xml
```diff
@@ -1921,286 +1921,578 @@ interpreted as follows:</para>
...
+</listitem>
+<listitem>
+<para>If the <code>filepath</code> is the string “//”, it has no
+scheme, no authority, and no drive letter. Its path is the
+string “/” and it is absolute and hierarchical.</para>
+</listitem>
+<listitem>
+<para>On a Windows system, if the <code>filepath</code> matches the
+regular expression “<code>^(file:/*)?([a-z]):(.*)</code>”, it is a
+“file” scheme URI. If the first match group begins “file:”, it is an
```
gimsieke, line 2052: begins → begins with

xproc/src/main/xml/specification.xml
```diff
@@ -1921,286 +1921,578 @@ interpreted as follows:</para>
...
+explicit file scheme URI, otherwise it is an implicit file scheme URI.
```
gimsieke, line 2053: Again I think that it hinges on the normalized `basedir` value. If its scheme is “ftp”, then the “u” in a `filepath` of “u:pass@ftp.acme.com/dir” shouldn’t be interpreted as a drive letter. 

Of course we can define `p:urify()` otherwise, but I think it will be more generic and consistent if file heuristics only apply to `filepath` if `basedir` is known to explicitly or implicitly represent a file URI.

gimsieke (changes requested):

xproc/src/main/xml/specification.xml
ndw: I think I can make the change you are suggesting, but I'm not sure it's worth the added complexity. As specified now,
```
p:urify("u:pass@ftp.acme.com/dir", "ftp://ftp.example.com/")
```
is an error (`err:XD0075` on Windows). If `u:` was not interpreted as a drive letter on Windows in this case, it would *still* be an error, `err:XD0077` just as it is on a non-Windows platform now.

Even if we imagine a yet more complex set of heuristics where the `u:` is not interpreted as a drive letter and not interpreted as a scheme and gets resolved against the base dir, it still wouldn't produce a result that's likely to be useful. The resolved URI would be:
```
ftp://ftp.example.com/u:pass@ftp.acme.come/dir
```
I can just about imagine a set of rules that would produce `ftp://u:pass@ftp.acme.com/dir` but only by adding a whole raft of special cases for different patterns and different schemes. That way lies madness, surely?

Is there a case where the current spec produces an error, the change suggested would remove the error without introducing a new one, **and** the result would be what the user was hoping for?

gimsieke (approved):

PR #1042, opened 2021-11-20
Attempt to resolve the select question in issue 1040

### LINKED ISSUE #1040: What happens if `select` on `p:with-input` isn't a string?
Martin Honnen [asks](https://xmlcom.slack.com/archives/C011NLXE4DU/p1636454612094200) on Slack:

> Trying to play a bit more with XProc 3 I wrote a pipeline with an empty sequence for the source port but then a p:identity step that tries to output a variable e.g.
```
<p:declare-step xmlns:p="http://www.w3.org/ns/xproc" version="3.0">
    <p:input port="source" sequence="true">
        <p:empty/>
    </p:input>
    <p:output port="result" sequence="true" serialization="map{ 'method' : 'json', 'indent' : false() }"/>
    <p:variable name="array-of-strings" as="array(xs:string)" select="['foo', 'bar']" xmlns:xs="http://www.w3.org/2001/XMLSchema"/>
    <p:identity>
        <p:with-input port="source" select="$array-of-strings"/>
    </p:identity>
</p:declare-step>
```
> Morgana XProcIII (latest beta) doesn't output anything as the result, XML Calabash 3 (latest preview) does output some kind of odd serialization of that JSON array. 

What do we think the right answer is? I started out thinking that an empty sequence was the right answer, but I've talked myself around to thinking that if `select` selects a variable, the actual document isn't relevant and what you get back is the selected variable.

But does that make sense?

ndw:
Close #1040 

This PR resolves the behavior of a `select` expression on `p:with-input` when the expression does not refer to the context item. (For example, if the `select` expression returns a static map or array).

On the surface, it introduces the slightly strange semantic that if a `select` expression is provided, but *nothing* appears on the input port, the expression is still evaluated once, against an empty document node.

Note that the context has to be an empty document or there will be all sorts of spurious errors. Consider:

```xml
<p:with-input port="source" select="//div"/>
```

If the expression was evaluated without a context item, that would raise an error which is surely not what the user expects.

Conversely, if we don't evaluate the expression at all when no documents appear on the input port, then this example:

```xml
<p:with-input port="source" select="map { 'a': 1 }"/>
```

produces no inputs to the step. And we've established (in #1040) that we *do* want the map to be sent to the step once.

### DIFF
--- a/xproc/src/main/examples/input-port.xml
+++ b/xproc/src/main/examples/input-port.xml
@@ -7,7 +7,8 @@
 </p:identity>
 <p:identity name="irrelevant">
 
-<p:with-input port="source" select="//html:div" xmlns:html="http://www.w3.org/1999/xhtml">
+<p:with-input xmlns:html="http://www.w3.org/1999/xhtml"
+              port="source" select="//html:div">
   <p:pipe step="origin" port="result"/>
 </p:with-input>
 
--- a/xproc/src/main/examples/input-select.xml
+++ b/xproc/src/main/examples/input-select.xml
@@ -2,7 +2,8 @@
 <p:output port="result"/>
 <p:identity>
 
-<p:with-input port="source" select="//html:div" xmlns:html="http://www.w3.org/1999/xhtml">
+<p:with-input xmlns:html="http://www.w3.org/1999/xhtml"
+              port="source" select="//html:div">
   <p:document href="http://example.org/input.html"/>
 </p:with-input>
 
--- a/xproc/src/main/xml/specification.xml
+++ b/xproc/src/main/xml/specification.xml
@@ -5292,25 +5292,22 @@ undefined.</error></para>
 </varlistentry>
 <varlistentry><term><tag class="attribute">select</tag></term>
 <listitem>
-<para>A <tag class="attribute">select</tag> expression
-<rfc2119>may</rfc2119> be provided with a connection.
-If a <tag class="attribute">select</tag> expression is specified,
-it will be applied to each item that appears on the connection. That is,
-for each item received, the
-<tag class="attribute">select</tag> expression will be evaluated with that
-item as the context item.
-If no items are received on the connection, the select
-expression will be evaluated once with an empty sequence as the
-context item.
-The sequence of items returned (the accumulated sequence of items, in the case where
-more than one item appears on the connection)
-by the expression is the input to the step.</para>
+<para>If a <tag class="attribute">select</tag> expression is
+specified, it is effectively a filter on the input. The expression
+will be evaluated once for each document that appears on the port,
+using that document as the context item. The result of evaluating the
+expression (on each document that appears, in the order they arrive)
+will be the sequence of items that the step receives on the
+port.</para>
 
 <para>The rules as stated in <xref
 linkend="creating-documents-from-xdm-step-results"/> will be applied to the members
 of this sequence and will turn these into separate documents. <error code="D0016">It is a
 <glossterm>dynamic error</glossterm> if the select expression on a <tag>p:input</tag> or
 <tag>p:with-input</tag> returns attribute nodes or function items.</error></para>
+
+<para>If no documents are received, the expression is not evaluated and the
+step receives no input on the port.</para>
 
 <para>For example:</para>
 
--- a/xproc/src/main/xml/specification.xml
+++ b/xproc/src/main/xml/specification.xml
@@ -5300,7 +5300,7 @@ for each item received, the
 <tag class="attribute">select</tag> expression will be evaluated with that
 item as the context item.
 If no items are received on the connection, the select
-expression will be evaluated once with an empty document node as the
+expression will be evaluated once with an empty sequence as the
 context item.
 The sequence of items returned (the accumulated sequence of items, in the case where
 more than one item appears on the connection)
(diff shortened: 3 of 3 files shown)


ndw:
My original PR suggested using an empty document node as the context for the `select` expression if there were no items on the port. I think that was an error. I've changed it to empty sequence. That may also be an error, but it's my current best suggestion.

ndw:
Achim and I have been discussing this in email. For context, here's part of that discussion (reproduced with Achim's consent):

Norm writes:

> Oh, I see. I thought that
>> <p:with-input select="//div">
>>   <p:empty />
>> </p:with-input>
>
> should fail, while 
>> <p:with-input select="5+3">
>>   <p:empty />
>> </p:with-input>
>
> should return 8. So the basic rule would be: If the XPath expression
> is based on the context item and there is no context item: Fail.

If that was the only way this could arise, I’d be fine with it. But I’m worried about this case:

```xml
  <p:pipeline>
    <p:output port="result" sequence="true"/>
    <p:option name="html"/>

    <p:if test="exists($html)">
      <p:http-request href="{$html}"/>
    </p:if>

    <p:identity>
      <p:with-input select="//div"/>
    </p:identity>
  </p:pipeline>
```

This is a somewhat contrived example, but I think the problem is clear. There’s no DRP for the p:if step, so if `$html` is not specified, or if the `p:http-request` step returns an empty sequence, then the input to `p:identity` will be the empty sequence. Then the attempt to get `//div` will fail where I think the very reasonable expectation of the pipeline author is that they’ll get an empty sequence.

> Would you suggest the same solution for
> <p:with-option name="opt" select="//div">
>   <p:empty />
> </p:with-option>

No, I don’t think so. We already have special rules for `p:with-option` if there are a sequence of documents, so it’s already different from `p:with-input`.

The semantics of `p:with-option` are clearer, and different, I think.

> I am not opposed to this change, but I am not sure it is necessary to make such a radical change. Aren't XPath people used to the fact that it is an error to rely on an absent context item?

Well, what do you think of my example above? I think XProc users expect `//div` to return the empty sequence not raise an exception. Maybe that case doesn’t really arise very often, but it feels like it might.

I think:

1. No context item is going to be surprising and unpleasant for pipeline authors. Do you think I’m wrong?
2. Empty document was a bad idea, we shouldn’t do that.
3. Will empty sequence work?
4. Or might it be better to say that the select expression is always applied to the context item(s) provided on the port and if there are no context items, the select expression is not evaluated.

But I’m troubled by the last option too. If you want to get an array
into a pipeline, this seems reasonable:

```xml
  <p:identity>
    <p:with-input select="[3,4]">
      <p:empty/>
    </p:with-input>
  </p:identity>
```

For this to produce an empty sequence is weird. But perhaps a lot easier to explain than `//div` raising an error!

All the options are bad! :-(

-------

Achim writes:

Ok, I see your point. Letting `@select` not fail but producing an empty sequence makes pipelines easier to write. In your example, I currently have to write:

```xml
 <p:pipeline>
    <p:output port="result" sequence="true"/>
    <p:option name="html"/>
 	
      <p:choose>
 	   <p:when test="exists($html)">
 	     <p:http-request href="{$html}"/>
 	   </p:when>
	     <p:otherwise>
               <p:identity>
		<p:with-input><failure /></p:with-input>
	       </p:identity>
	    </p:otherwise>
      </p:choose>
    <p:identity>
      <p:with-input select="//div"/>
    </p:identity>
  </p:pipeline>
```

And this would not cover the case, where `p:http-request` doesn't return a document.

Given your options (1)-(4), I think only (1) and (3) are reasonable: (4) would not solve the original problem and force pipeline authors to do things like

```xml
<p:identity>
	<p:with-input select="[4,5]">
		<i-am-completly-irrelevant />
	</p:with-input>
</p:identity>
```

From intuition I would opt for (1) to keep `p:with-input/@select` inline with `p:with-option/@select`. But I do clearly see the advantages of (3). Maybe I just have to get used to this.

gimsieke (approved):

gimsieke:
I think “empty sequence“ is a good solution, the least surprising/cumbersome under most circumstances.

xatapult (approved):
Elegant solution!

ndw:
Not as elegant as I hoped. Turns out "empty sequence" is synonymous with "context is absent".

That means an expression that does rely on the context will raise errors. So, this change would allow

```xml
<p:with-input select="[1,2,3]">
  <p:empty/>
</p:with-input>
```

to return `[1,2,3]`, but it would cause this

```xml
<p:with-input select="/div">
  <p:empty/>
</p:with-input>
```

to throw an exception. On balance, I'm not sure this change is a good idea. The rule that says the expression is applied to each item that appears on the port is neat and clean. If you want `[1,2,3]`, you have to do something like this:

```xml
<p:with-input select="[1,2,3]">
  <irrelevant-document/>
</p:with-input>
```

which is a bit ugly, but may be the least worst solution.

gimsieke:
If an empty document that the processor pulls out of its hat has already been rejected, then let’s go with what you just proposed (and which happens to be Achim’s original reading and implementation, I think). 

As a courtesy to users, a processor may produce a warning (that may be switched off) if a select expression meets an empty input.

ndw:
I've reworked it so that the `select` expression *only* applies to input documents.

gimsieke (approved):

PR #1046, opened 2021-11-26
Attempt to resolve issues 1030 and 1029

### LINKED ISSUE #1029: Default inputs
In 16.1, we say:

> On a `p:declare-step` for an atomic step, the `p:input` can only declare the input port. It is a static error (err:XS0042) to attempt to provide a connection for an input port on the declaration of an atomic step.

I don't think that's true. It's certainly at odds with this declaration:

```
<p:declare-step type="p:archive">
     <p:input port="source" primary="true" content-types="any" sequence="true"/>
     <p:input port="manifest" content-types="xml" sequence="true">
          <p:empty/>
     </p:input>
     ...
```

Which of the following is the best explanation for this apparent discrepancy:

1. It's fine, I'm just confused about what it means.
2. It used to be this way, but we changed our minds and decided to allow them, so this error should be removed.
3. The error is actually meant to say something else about the bindings.

### LINKED ISSUE #1030: Inconsistent use of “atomic step”
As described in https://github.com/xproc/3.0-specification/issues/1029#issuecomment-920186913, there are two different definitions of “atomic step“ in the 1.0 spec. 

The first definition is “step without subpipeline in its definition”, the second is “step without subpipeline in its invocation”.

In 3.0, we have mostly adopted the second definition.

However, there are remnants of the first, namely in:

---

### [§ 16.5. p:declare-step](https://spec.xproc.org/master/head/xproc/#p.declare-step)

> A p:declare-step provides the type and signature of a pipeline or an atomic step. Pipelines contain a subpipeline which defines what the declared step does. Atomic steps have an implementation defined elsewhere in some other way.

and

#### [§ 16.5.2. Declaring atomic steps](https://spec.xproc.org/master/head/xproc/#declare-atomic-steps)

> The distinction between an atomic step declaration and a pipeline declaration hinges on the presence or absence of a subpipeline. An atomic step declaration does not have a subpipeline. 

> Atomic step declarations may not import other pipelines or functions, may not declare static options, and may not declare additional steps. In other words, the content of an atomic step declaration consists exclusively of p:input, p:output, and p:option elements.

In addition, the two static errors [`err:XS0042`](https://spec.xproc.org/master/head/xproc/#err.inline.S0042) and [`err:XS0029`](https://spec.xproc.org/master/head/xproc/#err.inline.S0029) also use “atomic step” in its first sense.

---

This is not how we defined atomic steps in

### [§ 15.8. Atomic Steps](https://spec.xproc.org/master/head/xproc/#p.atomic)

> [Definition: An atomic step is a step that does not contain a subpipline when it is invoked.]

I propose that we use a different term than “atomic step” for the first meaning, “step without subpipeline in its declaration,” and replace it in § 16.5 and in both error messages (in case they survive #1029).

ndw:
I took on board @gimsieke 's suggestion that the spec would be clearer if it distinguished between two different meanings of "atomic step", as he outlines in #1030. I decided to define a new term "external step" for steps that do not contain a subpipeline.

I've redrafted 16.5 accordingly and I do think it's an improvement. I think it also clarifies `err:XS0029`.

In #1029 there was a proposal to remove `err:XS0042`. That's already been done by some other PR. I assume that was accidental, but I think it's clearly what we needed to do so I'm just going to let it stand.

Close #1029 
Close #1030

### DIFF
--- a/xproc/src/main/xml/specification.xml
+++ b/xproc/src/main/xml/specification.xml
@@ -5558,10 +5558,9 @@ parameters</link>.
 </varlistentry>
 </variablelist>
 
-
 <para><error code="S0029">It is a <glossterm>static error</glossterm>
 to specify a connection for a <tag>p:output</tag> inside a
-<tag>p:declare-step</tag> for an atomic step.</error></para>
+<tag>p:declare-step</tag> for an <glossterm>external step</glossterm>.</error></para>
 
 <para>If a connection is provided for a <tag>p:output</tag>, documents
 are <emphasis>read from</emphasis> that connection and those documents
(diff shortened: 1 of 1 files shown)


xproc/src/main/xml/specification.xml
```diff
@@ -6195,10 +6194,20 @@ specify a value for an option that is declared static.</error>
...
-other way.</para>
+step does.
+<termdef xml:id="dt-external-step">An <firstterm>external step</firstterm>
+is one supported by the implementation, but which has no exposed description
+of what it does.</termdef> 
+</para>
+
+<para>The standard XProc atomic steps (<tag>p:add-attribute</tag>,
+<tag>p:add-xml-base</tag> …) are all external steps.
+<impl>Whether or not an implementation allows users to provide thier own
```
gimsieke, line 6207: their

gimsieke (approved):
approve if typo is corrected

PR #1054, opened 2021-12-03
Attempt to describe deadlock in use-when expressions

### LINKED ISSUE #1051: Deadlock conditions in `use-when` aren't discussed
Consider:

```xml
<p:declare-step type="cx:lib2d1" use-when="not(p:step-available('cx:lib2d2'))">
  <p:output port="result"/>
  <p:identity>
    <p:with-input><doc>lib1s1</doc></p:with-input>
  </p:identity>
</p:declare-step>

<p:declare-step type="cx:lib2d2" use-when="not(p:step-available('cx:lib2d1'))">
  <p:output port="result"/>
  <p:identity>
    <p:with-input><doc>lib1s1</doc></p:with-input>
  </p:identity>
</p:declare-step>
```

It's relatively difficult to create such a deadlock. You have to put the `use-when` expression on a `p:declare-step` and it has to include a call to `p:step-available`. There are a number of options for resolving them, but I think the simplest thing to do for 3.0 is make this a static error.

ndw:
Close #1051

### DIFF
--- a/xproc/src/main/xml/specification.xml
+++ b/xproc/src/main/xml/specification.xml
@@ -3768,6 +3768,39 @@ linkend="initiating">static analysis</link> of the pipeline.</para>
 processing occurs after XML parsing and has no effect on well-formedness
 or validation errors which will be reported in the usual way.</para>
 </note>
+
+<para>Deadlock situations can arise if two or more 
+<tag class="attribute">[p:]use-when</tag> expressions depend on each other.
+Consider, for example:</para>
+
+<programlisting language="xml"><![CDATA[<p:declare-step type="ex:A" use-when="p:step-available('ex:B')">
+  …
+</p:declare-step>
+
+<p:declare-step type="ex:B" use-when="p:step-available('ex:A')">
+  …
+</p:declare-step>]]></programlisting>
+
+<para>It is not possible for the processor to determine if
+<code>ex:A</code> should be declared without first determining if
+<code>ex:B</code> should be declared, and vice versa.</para>
+
+<para><error code="S0115">It is a <glossterm>static error</glossterm>
+if two or more elements are contained within a deadlocked network of
+<tag class="attribute">[p:]use-when</tag> expressions.</error> In
+order to avoid deadlock, there must exist an order in which every
+expression can be resolved without reference to an expression that
+occurs after it in the ordering.</para>
+
+<para>Processors may be required to evaluate expressions in an
+arbitrary order, but they are not required to solve a set of linear
+equations simultanously. So, while “declare both <code>ex:A</code> and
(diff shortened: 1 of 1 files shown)


xproc/src/main/xml/specification.xml
```diff
@@ -3768,6 +3768,39 @@ linkend="initiating">static analysis</link> of the pipeline.</para>
...
+<para><error code="S0115">It is a <glossterm>static error</glossterm>
+if two or more elements are contained within a deadlocked network of
+<tag class="attribute">[p:]use-when</tag> expressions.</error> In
+order to avoid deadlock, there must exist an order in which every
+expression can be resolved without reference to an expression that
+occurs after it in the ordering.</para>
+
+<para>Processors may be required to evaluate expressions in an
+arbitrary order, but they are not required to solve a set of linear
+equations simultanously. So, while “declare both <code>ex:A</code> and
```
gimsieke, line 3797: simultaneously

xproc/src/main/xml/specification.xml
```diff
@@ -3768,6 +3768,39 @@ linkend="initiating">static analysis</link> of the pipeline.</para>
...
+<code>ex:B</code>” is a valid solution to the example above,
+conformation processors are not required (or allowed) to find it
```
gimsieke, line 3799: conformant

xproc/src/main/xml/specification.xml
```diff
@@ -3768,6 +3768,39 @@ linkend="initiating">static analysis</link> of the pipeline.</para>
...
+<code>ex:B</code> should be declared, and vice versa.</para>
+
+<para><error code="S0115">It is a <glossterm>static error</glossterm>
+if two or more elements are contained within a deadlocked network of
+<tag class="attribute">[p:]use-when</tag> expressions.</error> In
+order to avoid deadlock, there must exist an order in which every
+expression can be resolved without reference to an expression that
+occurs after it in the ordering.</para>
+
+<para>Processors may be required to evaluate expressions in an
```
gimsieke, line 3795: may be required → may be allowed?

gimsieke (changes requested):

xproc/src/main/xml/specification.xml
ndw: 👍

xproc/src/main/xml/specification.xml
ndw: 👍

xproc/src/main/xml/specification.xml
ndw: I think required is correct. The processor has to determine an order that works, and whether or not there is one, and then evaluate the expressions in that order.

gimsieke (approved):

PR #1060, opened 2022-04-27
Attempt to fix #1056

### LINKED ISSUE #1056: The spec is contradictory about the content of p:option
In 7.1, we have an example of a static option that contains a `p:document`. But in 16.4.2, we define the content model of `p:option` as empty. 

My guess is that we intended to allow `p:document` and `p:inline` (and implicit inlines) in *static* `p:option` elements.

But we haven't written it that way yet.

xml-project:
I think this would be the easiest way to fix the problem in prose. Does it make sense to anyone?


gimsieke (approved):
Seems like a good fix

ndw (approved):

PR #1067, opened 2022-06-13
Attempt to fix issue 1066

### LINKED ISSUE #1066: Legacy example for p:http-request?
In [15.7.4](https://spec.xproc.org/master/head/xproc/#example-try) I think the example uses c:request/c:body concept from XProc 1.0 to express the content of an HTTP-request.
I think the XProc 3.0 way should be:
````
<p:http-request method="post" href="http://example.com/form-action">
   <p:with-input>
      <p:inline content-type="application/x-www-form-urlencoded">name=W3C&amp;spec=XProc</p:inline>
   </p:with-input>
</p:http-request>
````
Did I miss something?

xml-project:
Should fix #1066

### DIFF
--- a/xproc/src/main/examples/trycatch.xml
+++ b/xproc/src/main/examples/trycatch.xml
@@ -4,13 +4,9 @@
 <p:output port="result"/>
 
 <p:try>
-  <p:http-request>
-    <p:with-input port="source">
-      <p:inline>
-	<c:request method="post" href="http://example.com/form-action">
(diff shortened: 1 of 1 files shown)


gimsieke:
Uh, I notice that I wouldn’t even know how to form such a request. In my opinion the standard steps spec should also feature an example or two for `p:http-request`. With respect to approving this PR, I politely ask Norm or Erik to step in.

PR #1090, opened 2024-04-12
Attempt to clarify the semantics of static options

### LINKED ISSUE #1085: Clarify the scope of static options
Since we're going to be considering errata later this week, I don't think the description of static options is precise enough. Consider this short paragraph from the spec:

> XProc defines a single, global scope for static options. Every static option must have exactly one in-scope declaration.

I don't think those two sentences are coherent. Consider:

```
p:declare-step name="main"

  p:declare-step name="step1"
    p:import href="lib1"
  ...

  p:declare-step name="step2"
    p:import href="lib2"
  ...
...
```

where `lib1` declares

```
  <p:option name="sopt" select="xs:date('2024-04-09')" as="xs:date" static="true"/>
```

and `lib2` declares

```
  <p:option name="sopt" select="'floor wax'" as="xs:string" static="true"/>
```

There's only one in-scope declaration for each of those options, but they're clearly incompatible if there's a single, global scope.

I think it's probably sufficient to just remove the reference to a single global scope. We don't allow redeclaration of static options, so if they're in the same scope, that's an error. If they're in different scopes, I don't think there's anything ambiguous about them. In `step1`, `$sopt` is a date, in `step2`, `$sopt` is a string.

If the user attempts to define a static value for `$sopt` at compile time, we say:

> There may be an [implementation-defined](https://spec.xproc.org/master/head/xproc/#dt-implementation-defined) mechanism for providing default values for static [p:option](https://spec.xproc.org/master/head/xproc/#p.option)s. If such a mechanism exists, the values provided must match the sequence type declared for the option, if such a declaration exists.

If the provided value doesn't satisfy the declared type of `$sopt` (wherever that declaration occurs), that's a static error. In this example, I don't think there is any such value.

### LINKED ISSUE #1086: Do private static options make any sense?
It strikes me as weird that some deeply nested library could define

```
<p:option name="sopt" select="7" static="true" visibility="private"/>
```

And now it's an error for my pipeline to define a static option named `sopt` (either at all or that isn't a numeric, depending on how we resolve #1085).

It's also weird that an implementation might provide an API for defining the value of an otherwise invisible option.

(For what it's worth, I think the right answer here is "yes, it's a little weird, but that's the way it works because making it work some other way would be differently and more profoundly weird. Best practice is to define private, static options in a namespace that assures they won't collide with the names of static options used in other libraries." But if that's what we think, we should say that.)

ndw:
Close #1085 
Close #1086 

This is also mostly editorial. I've made `err:XS0088` apply to static options that shadow other static options. We didn't previously identify an explicit error code in this section, but I think that's a reasonble choice.

### DIFF
--- a/xproc/src/main/xml/specification.xml
+++ b/xproc/src/main/xml/specification.xml
@@ -3025,15 +3025,34 @@ Variable names are always expressed as
     <section xml:id="statics">
       <title>Static Options</title>
 
-      <para>A <tag>p:option</tag> may be declared “static”; options declared within a <tag>p:library</tag> <rfc2119>must</rfc2119>
-      be static.</para>
-      <para><error code="S0109">It is a <glossterm>static error</glossterm> if options that are the direct children of 
-        <tag>p:library</tag> are not declared “static”</error>.</para>
-      <para>The values of static options are computed during <link linkend="initiating">static
-          analysis</link>.</para>
+      <para>A <tag>p:option</tag> may be declared “static”; options declared
+      within a <tag>p:library</tag> <rfc2119>must</rfc2119> be static.
+      <error code="S0109">It is a <glossterm>static error</glossterm> if
+      options that are the direct children of <tag>p:library</tag> are not
+      declared “static”</error>.</para>
 
-      <para>XProc defines a single, global scope for static options. Every static option must have exactly one
-      in-scope declaration.</para>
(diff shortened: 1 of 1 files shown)


xml-project (approved):
As XS0088 is also used for a variable shadowing a static option, maybe we should change the error text:
"It is a static error ... of a static option or a variable ..."

PR #1092, opened 2024-04-14
Make the default XPath version implementation defined

ndw:
Closes issue 39 in the vnext repository.

### DIFF
--- a/xproc/src/main/xml/specification.xml
+++ b/xproc/src/main/xml/specification.xml
@@ -6354,13 +6354,14 @@ If the attribute is not specified, the value
 <rfc2119>must</rfc2119> be used to evaluate XPath expressions subject
 to the constraints outlined in <xref linkend="xpath-context"/>.
 <error code="S0110">It is a <glossterm>static error</glossterm> if
-the requested XPath version is less than “<literal>3.1</literal>”</error>.</para>
+the requested XPath version is less than “<literal>3.1</literal>”</error> or is
+not supported by the processor.</para>
 <para><impl>If a pipeline does not request a specific XPath version, the version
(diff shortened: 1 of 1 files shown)


xproc/src/main/xml/specification.xml
```diff
@@ -6352,10 +6352,17 @@ If the attribute is not specified, the value
...
-assumed. <error code="S0110">It is a <glossterm>static error</glossterm> if
+to the constraints outlined in <xref linkend="xpath-context"/>.
+<error code="S0110">It is a <glossterm>static error</glossterm> if
 the requested XPath version is less than “<literal>3.1</literal>”</error>.</para>
+<para><impl>If a pipeline does not request a specific XPath version, the version
+used is <glossterm>implementation-defined</glossterm>.</impl> <impl>If different
+pipelines or libraries declare different XPath versions, it is
+<glossterm>implementation-defined</glossterm> how those conflicts are resolved.</impl>
+An implementation might use different versions for different pipelines, or it might
+use the same version for all pipelines. <impl>If an implementation elects to us the
```
gimsieke, line 6363: us → use

gimsieke (changes requested):
minor typo

xml-project:
I think this is a good solution for the inheritance problem. But I am not sure how it covers the problem when the pipeline is asking for an XPath version not supported by the processor. Does this PR suggest the reaction as implementation-defined?

ndw:
Fixed the typo. Extended `err:XS0110` to cover versions less than 3.1 or other versions unsupported by the processor.

gimsieke (approved):

PR #1111, opened 2024-08-28
Attempt to clarify p:import-functions

### LINKED ISSUE #1110: Clarify p:import-functions
When are functions loaded? Are they accessible in `[p:]use-when` expressions during static analysis?

ndw:
Fix #1110

### DIFF
--- a/xproc/src/main/xml/specification.xml
+++ b/xproc/src/main/xml/specification.xml
@@ -3578,10 +3578,22 @@ library <rfc2119>must</rfc2119> have unique names: it is a
 <glossterm>static error</glossterm> if any static option name is
 declared more than once in the same scope.</error></para>
 </section>
-      
+
+<section>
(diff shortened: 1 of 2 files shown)


gimsieke (approved):

xml-project (changes requested):
I am sorry to disagree with one aspect, but I think that functions imported in a p:library should not be automatically visible in a pipeline importing that library. As far as I can see, this break the current scoping rules for import.
If I want to use a function in a step, a p:import-function should be necessary to avoid scoping problems in the importing step.

ndw:
Yes. I think you're absolutely right. I did note in the issue that I wasn't sure what the right answer was. Thanks for telling me :-)

xml-project:
@ndw 

> I did note in the issue that I wasn't sure what the right answer was.

Sorry, I missed that question.

PR #1120, opened 2024-12-14
Describe negated shortcut content types

### LINKED ISSUE #1115: Clarify that negated content type shortcuts are allowed
The current wording of [3.4 Specifying content types](https://spec.xproc.org/3.0/xproc/#specified-content-types) suggests that `-application/xml` is allowed, but `-xml` is not. That was perhaps the case at one time, but we clearly decided it was okay when we specified `p:invisible-xml`. 

We should clarify that shortcuts can be negated.

We should also consider if there are any additional error conditions we can capture. `-any` seems pointless. But so does `application/xml text/xml -xml`. What's not clear on five minutes thought is whether or not it's practical to detect this error statically.

ndw:
Fix #1115 

Note how I've proposed to handle negation of exclusions.

### DIFF
--- a/xproc/src/main/xml/specification.xml
+++ b/xproc/src/main/xml/specification.xml
@@ -811,6 +811,12 @@ The following shortcuts <rfc2119>must</rfc2119> be supported by the processor:</
 </varlistentry>
 </variablelist>
 </listitem>
+<listitem>
+<para>A shortcut form preceded by a minus sign (“-”) indicates that the specified types are forbidden.
+When expanding a negated shortcut, any forbidden types in the expansion are ignored. In other
+words, <literal>-xml</literal> expands to “<literal>-application/xml -text/xml -*/*+xml</literal>”,
+the exclusion on HTML documents is ignored.</para>
+</listitem>
 </itemizedlist>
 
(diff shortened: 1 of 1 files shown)


gimsieke (approved):
makes sense

xml-project (commented):
I am not sure, I got your example right.
The spec says:

> The document is considered acceptable if and only if it matches at least one acceptable content type and the last content type that matched was not forbidden.

Now consider a port defined with "-xml". A document marked with "text/plain" should be accepted, but: It does not match any acceptable content type.
What do I miss?

gimsieke:
Maybe we need to say that if the list only expands to exclusions, `*/*` will be prepended?

ndw:
It's an interesting question. I think that just "doesn't match at least one acceptable type". Just saying `-application/xml` wouldn't match `text/plain` either. If you don't have any positive matches, I think that's just not going to work. When we actually use it in the Invisible XML step, we write:
```
any -xml -html
```

ndw:
There are already ways to make a port that doesn't accept any documents, I don't think we have to call attention to `-any`, but we could.

xml-project:
@ndw Ah, ok! I obviously missed the main point. Now your proposal makes totally sense to me. Lets see how often forget about this point when writing pipelines.

xml-project (approved):

PR #1121, opened 2024-12-14
Attempt to finesse 3.0 vs 3.1

### LINKED ISSUE #1116: What version(s), exactly?
The section [Versioning Considerations](https://spec.xproc.org/master/head/xproc/#versioning-considerations) in the 3.1 specification is a hot mess. It still very explicitly says "3.0" which either needs to change or be clarified. Also, we say:

> XProc 3.0 takes a draconian approach to versioning. It is a [static error](https://spec.xproc.org/3.0/xproc/#dt-static-error) ([err:XS0060](https://spec.xproc.org/3.0/xproc/#err.S0060)) if the processor encounters an explicit request for a version of the language other than “3.0”.

What are we doing about that? I think we could say:

1. 3.1 doesn't change the version of pipeline documents, it's a minor revision, you must still use `"3.0"`
2. Or we could say that `"3.1"` is allowed
3. Or we could say that `"3.1"` is required.

I don't think option 3 is a good idea. I could live with 1 or 2. I'd want to make it implementation defined, however, if a 3.1 processor must handle 3.0 pipelines in some backwards compatible way (using the old, 3.0 library declarations for example). I'm not eager to do that.

ndw:
Fix #1116

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -132,7 +132,8 @@ spec_schemas.dependsOn download_core30_rng
 
 def schemaList = ["xproc.rnc", "xproc.rng",
                   "xproc10.rnc", "xproc10.rng",
-                  "xproc30.rnc", "xproc30.rng"]
+                  "xproc30.rnc", "xproc30.rng",
+                  "xproc31.rnc", "xproc31.rng"]
 
 schemaList.each { schema ->
   def taskname = "download_" + schema.replace(".", "_")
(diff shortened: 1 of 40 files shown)


gimsieke (approved):

ndw:
@gimsieke approved this a while ago, but I pushed a bunch of new changes. It would be nice to get confirmation again from Gerrit or @xml-project  or @xatapult

gimsieke:
I’d still approve it, but maybe Achim or Erik want to give their approval, too.

ndw:
I'm happy you took a second look. I'll give it a day or two and merge it if no one objects (or if anyone else explicitly approves).

PR #1147, opened 2025-04-24
Define support for query and fragment portions of a URI in p:urify()

### LINKED ISSUE #1145: The p:urify() function makes no provision for queries or fragment identifiers
I can't decide if the `p:urify()` function is intentionally or accidentally oblivious to queries and fragment identifiers. I think we should say which.

If it's intentional, I think users should be warned about these results

```
p:urify("#fragid", "file:///path/to/thing") ==> "file:///path/to/%23fragid"
p:urify("x#fragid", "file:///path/to/thing") ==> "file:///path/to/x%23fragid"
p:urify("?x#fragid", "file:///path/to/thing") ==> "file:///path/to/%3Fx%23fragid"
```

I was quite surprised to discover them. I'm having a hard time persuading myself that this can be correct though. I think in the days before RFC 3986, there was some distinction between URI and URI reference and fragment identifiers were only part of a URI reference. But RFC 3986 seems to be pretty definitive:

```
URI         = scheme ":" hier-part [ "?" query ] [ "#" fragment ]
```

If we wanted to fix this, I think this would work:

1. If the `filepath` is hierarchical and contains a query or a fragment identifier, the query or fragment identifier is stripped off as a `suffix` and set aside.
2. If the remaining `filepath` is the empty string, then we need a special case so that the last path segment doesn't get stripped off.
3. The `suffix`, if it exists, is appended to the result.

With those changes, you get:

```
p:urify("#fragid", "file:///path/to/thing") ==> "file:///path/to/thing#fragid"
p:urify("x#fragid", "file:///path/to/thing") ==> "file:///path/to/x#fragid"
p:urify("?x#fragid", "file:///path/to/thing") ==> "file:///path/to/thing?x#fragid"
```

Those results are more consistent with what my expectations were.

And, regardless of the answer we choose, we need tests in the test suite for these cases.

ndw:
Fix #1145

Note: This is a speculative PR. The editorial team hasn’t agreed to this change, but if they do agree, I believe that this PR implements the proposed ammendments.

### DIFF
--- a/xproc/src/main/xml/specification.xml
+++ b/xproc/src/main/xml/specification.xml
@@ -2023,8 +2023,8 @@ analysis.</para>
 <para>The heuristics that <function>p:urify</function> performs occur
 in two stages: first, the input string is analyzed to identify its
 features, then these features are used to construct a final URI
-string. An additional “fixup” process may be performed on the path
-portion of the URI.</para>
+string. An additional “fixup” process may be performed on
+portions of the URI.</para>
 
 <para>To make the operation of the <function>p:urify</function>
 function easier to understand, the description that follows is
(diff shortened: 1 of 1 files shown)


gimsieke (approved):

xml-project:
@ndw Sorry for coming late.

xml-project (approved):

ndw:
Nothing to be sorry about. I'm just glad we agree!

